PR #89, opened 2019-01-17
Added markup and fileSize validation for incoming SLD urls

simonvanbennekom:
*removed a double check for empty SLD param
*Modified WMSSLD tests

Added markup and fileSize validation for incoming SLD urls

maxFileSize for incomming slds can be set in CSLD.h (MAX_FILE_SIZE_ALLOWED)
Changed some things in CHTTPTools to support maxFileSize and added readable error message


maartenplieger:
Great work!

PR #91, opened 2019-02-22
CConvertEProfile now also accepts files which have a global attribute…

saskiawagenaar:
… featureType:profile

### DIFF
--- a/adagucserverEC/CConvertEProfile.cpp
+++ b/adagucserverEC/CConvertEProfile.cpp
@@ -33,29 +33,42 @@ const char *CConvertEProfile::className="CConvertEProfile";
 
 
 /**
- * This function adjusts the cdfObject by creating virtual 2D variables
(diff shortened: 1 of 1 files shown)


saskiawagenaar:
Moet ik ook nog ergens documentatie bijwerken?

maartenplieger:
The following code

if( cdfObject->getAttribute("featureType")->toString().equalsIgnoreCase("profile")==false) 

will throw an exception if featureType is not found. That means that this class will never make it to the original eProfile checks if a file without a featureType is given. I think the original eProfile files don't contain a featureType attribute, so they will not be picked up anymore.

Solution: Check needs to be done separately for the new files and the original eProfile files. If one of the two is succesful, continue.

saskiawagenaar:
Good point, I forgot to test with the original E-profile file.
I'll fix it and test if the solution works with both the original files and the new files.

PR #115, opened 2019-08-27
Now checking on dimension names instead of dimension length to determ…

saskiawagenaar:
…ine if it is an E-Profile variable.


maartenplieger:
Those checks where for some corner cases, but if you don't encounter those in your application this is good.

PR #125, opened 2020-01-14
Add rad and radian detection

maartenplieger:
image

Added support for radian units (For @magau)


magau:
Thank you so much @maartenplieger! ;)

PR #137, opened 2020-07-10
Agaguc server ubuntu20

FStriewski:
Changes:
- Add updated deploy script for Ubuntu 20.04
- Add macros to prevent build issue with proj_api


maartenplieger:
Already there.

PR #139, opened 2020-08-11
Upgrade Python

FStriewski:
**Scope:**
Dockerfiles and Python tests are currently using version 2 and needed migration to Python 3.

**Changes:**
- Updated docker files to install Python 3, update libraries where necessary.
- Update unit tests to use Python 3 syntax.

**Test:**
You can run `runtests.sh` from the main folder to trigger all unit tests.
Check if the libraries are installed fine or if any requirements are missing, when using pip.


maartenplieger:
All tested, looks good!

PR #143, opened 2020-09-15
Inlinelegendselection

maartenplieger:
The showlegend parameter to the getmap URL has been extended. Previousely it accepted true or false, to either show or not to show an inline legend in the getmap request.

The extension allows for a comma separated list with layernames. These will be rendered in the GetMap accordingly.


FStriewski:
Nice!

PR #181, opened 2021-11-05
Fixed negative values for CSV reader

maartenplieger:
Negative values for CSV should now be detected as int and float. Colour scaling over these values should now work. 

Added some tests. Pushed to openearth/adaguc-server:2.6.1


hclasses/CTString.cpp
```diff
@@ -657,72 +658,70 @@ namespace CT{
...
-    if (inputLength == 1 && inputStr[0] == '.') {
+    const char *inputStr = this->c_str();
+    /*check size */
+    if (this->empty() || inputLength > 39)
+    {
       return false;
     }
-    if (inputLength == 3 && this->equals("NaN")) {
+    /* NaN is a "number"...in this context */
+    if (inputLength == 3 && this->equals("NaN"))
```
ernstdevreede: Check for length 3 is not necessary: is implicit in equals("NaN")

hclasses/CTString.cpp
```diff
@@ -657,72 +658,70 @@ namespace CT{
...
-    if (inputLength == 1 && inputStr[0] == '.') {
+    const char *inputStr = this->c_str();
+    /*check size */
+    if (this->empty() || inputLength > 12)
+    {
       return false;
     }
-    if (inputLength == 3 && this->equals("NaN")) {
+    /* NaN is a "int"...in this context */
+    if (inputLength == 3 && this->equals("NaN"))
```
ernstdevreede: NaN is not possible for ints

belentorrente:
In one place there is a check that  `if (this->empty() || inputLength > 12) {return false; }` . In another one` if (this->empty() || inputLength > 39) {return false; } `. Shouldn't these limits be coded somewhere so we know exactly what they mean? For example I imagine the 12 refers to a date like [2019-07-28],  so having a maxDateLength defined somewhere would be nice.

Issue #180, opened 2021-11-04
CSV  Not reading negative floats or integers


EduardoCastanho:
hi 
I'm trying to display a csv with negative temperatures  something like this :
```
lat,lon,tn
-8,38,-1.02
-8.5,38,2.02
(...)
```
and my var tn is classified as a string where should be a float, not allowing a color scale in adaguc 
Looking in the adaguc code the lines that classify the type :  
https://github.com/KNMI/adaguc-server/blob/d1699f2564ec4af8a69454e43ad6122754188d06/CCDFDataModel/CCDFCSVReader.cpp#L271-L277
and checking isFloat() function:
https://github.com/KNMI/adaguc-server/blob/d1699f2564ec4af8a69454e43ad6122754188d06/hclasses/CTString.cpp#L704-L727

I notice that the '-' sign is not in the allowed chars, therefore  not  classifying the negative float correctly .

I also notice the function is not checking that a float can only have one decimal separator (.) allowing '1.2.3' to be a "valid" float .

Thanks, 
Cheers

maartenplieger:
Thank you very much for the issue. That should be fixed soon. We will let you know.

maartenplieger:
Maybe it is solved in https://github.com/KNMI/adaguc-server/pull/181 ?

EduardoCastanho:
Hi ! 
Your change "fix" the negative number issue, but cause a new issue, 
for example: 
```
lat,lon,ffmaxmin,date_ffmaxmin,station_id
39.76362,-9.03059,-8.2,2019-07-28,721
39.76362,-9.03059,-8.2,2019-07-28,721
32.63655,-16.93559,0.0,2021-08-31,971
```
The date  column `date_ffmaxmin` will be treated like a float:
 
![dates_as floats](https://user-images.githubusercontent.com/11145763/140777642-9c54c31f-0d66-4bd6-8422-5ceaba9393bd.png)

and the multiple decimal separator  still will be accepted as a float 

I submit a pull request #182 I hope this solve the issue 


cheers

maartenplieger:
I think it is solved now. Thanks for your help!

PR #182, opened 2021-11-08
Alternative implementation with Regex

EduardoCastanho:
fix for the issue #180


maartenplieger:
Hi Eduardo, 

Very nice!

PR #183, opened 2021-11-16
replace environment variables defined in ADAGUC_EENV

kktuax:
It would be nice to have support for additional environment variables in the config files. We are planning to use this in order to configure styles, for instance:

```xml
<Style name="feat-highlight">
  <Legend fixed="true">feat-highlight-black</Legend>
  <FeatureInterval match="^{FEAT_MATCH}$" matchid="{FEAT_MATCH_ID}" fillcolor="#CCCCCC" label="unknown" />
  <RenderMethod>nearest</RenderMethod>
  <NameMapping name="nearest" title="highlight" />
</Style>
```

Could be used passing as environment variables:

```
FEAT_MATCH=168
FEAT_MATCH_ID=id
ADAGUC_EENV=FEAT_MATCH,FEAT_MATCH_ID
```

### DIFF
--- a/adagucserverEC/CServerParams.cpp
+++ b/adagucserverEC/CServerParams.cpp
@@ -25,6 +25,9 @@
 
 #include "CServerParams.h"
 #include "CStopWatch.h"
+#include <string>
+#include <vector>
+#include <sstream>
 const char *CServerParams::className="CServerParams";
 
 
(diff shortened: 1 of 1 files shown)


maartenplieger:
This feature is now available via the https://github.com/KNMI/adaguc-server/blob/master/doc/configuration/Environment.md setting.

PR #219, opened 2022-07-08
2262 quantize product times on ingest

ernstdevreede:
This change make sit possible for ADAGUC to apply quantizemthod/quantizeperiod attributes to product times on ingest (instead of only on TIME= request values as it was originally).
This makes it possible to quantize for example NOAA GOES16 product times, which is useful because GOES16 product times can vary a few seconds.

Should be enabled by configurig the time dimension, for example like this:
`   <Dimension name="time" quantizeperiod="PT10M" quantizemethod="round">time</Dimension>
`
quantizemethod can be `low`,` high` or `round` (`round` is the default).

Behaviour was also fixed for `quantizemethod="high"`: if a value is exactly on a quantizeperiod step, it is **not** rounded up. (For example with `quantizemethod="high" quantizeperiod=PT10M` the timestamp 2022-07-08T12:00:00Z yields 2022-07-08T12:00:00Z


CCDFDataModel/CTime.cpp
```diff
@@ -911,26 +911,35 @@ double CTime::quantizeTimeToISO8601(double offsetOrig, CT::string period, CT::st
...
       hmsPart.replaceSelf("M", "");
       int M = hmsPart.toInt();
       int origM = date.minute;
       date.second = 0;
-      date.minute = origM - origM % M;
+      int restM = origM % M;
+      date.minute = origM - restM;
       offsetLow = thisTime->dateToOffset(date);
-      date.minute = date.minute + M;
+      if (restM != 0) {
```
belentorrente: Is there a reason why it's `>0` for the restH comparison, but `!=0` here?

CCDFDataModel/CTime.cpp
```diff
@@ -972,9 +987,12 @@ double CTime::quantizeTimeToISO8601(double offsetOrig, CT::string period, CT::st
       hmsPart.replaceSelf("D", "");
       int D = hmsPart.toInt();
       int origD = date.day;
-      date.day = origD - origD % D;
+      int restD = origD % D;
```
belentorrente, line 990: There is a pattern with these operations, maybe you could make a 3-line aux function and apply it to every number you quantize.

CCDFDataModel/CTime.cpp
ernstdevreede: Changed this to >0 for consistence

CCDFDataModel/CTime.cpp
ernstdevreede: I don't see how this would make the code significantly clearer, so I don't feel it is necessary to make use a function here. It will probably not even save lines in the code.

PR #222, opened 2022-07-26
Fix log labels and test data

belentorrente:
When using a log scale, a series of standard ticks will be used for every power of 10: 2, 5, 10. In addition to that, a tick for min and another tick for max are displayed. 

image


belentorrente:
This is what it would look like using 30 instead:

test_WMSGetLegendGraphic_log10

Using 25 looks less intuitive.

maartenplieger (approved):

PR #224, opened 2022-09-06
adding support for projection_x/y_angular coordinates

lllliso:
From CF 1.9 onwards, the x (abscissa) and y (ordinate) rectangular coordinates are identified by the standard_name attribute values projection_x_angular_coordinate and projection_y_angular_coordinate, respectively. It is assumed that the y-axis is aligned to the Earth’s N/S axis, whereas the x-axis aligns with the E/W axis. CF specified the standard names projection_x_coordinate and projection_y_coordinate for these coordinates prior to version 1.9, but that use is deprecated. In the case of this projection, the projection coordinates are the scanning angle of the satellite instrument.
The standard units for this coordinates are "radian". 
A procedure has been coded to deal with both use cases of the "geostationary" projection. In the case of use of projection_x/y_angular coordinates x & y are not converted to meters and the proj internal string is overwritten to give the expected results. The changes are only applied on the c++ part...


maartenplieger:
Thanks for your PR.

Can you please make sure that the command
```
bash runtests.sh
```
Gives OK for all 88 tests?

adagucserverEC/CDataReader.cpp
```diff
@@ -491,64 +491,191 @@ int CDataReader::parseDimensions(CDataSource *dataSource, int mode, int x, int y
...
-              }
-            }
-          }
-        }
-      }
-    }
+    // coordinates can be converted to (k)m. In the geostationary projection under CF-1.9
+    // specifications, the x y values are kept, in this case the proj string is overwrote internally.
+    
+    CDF::Attribute *X_standard_name = dataSource->varX->getAttributeNE("standard_name");
```
maartenplieger, line 497: I think a lot of the code is now not functional if no standard_name is given. Can we make the if/else a bit less specific? I think many functional tests are failing because of this. If you need suggestions/help let me know.

lllliso:
OK I will run the tests

lllliso:
I will review the code next week, but you are true if no standard_name it avoids totally the lines. I will leave as the only execption the projection_x_angular_coordinate. And run the tests sorry for so much tests failed :)

maartenplieger:
No worries, if you need help let me know

PR #225, opened 2022-09-15
Added Point support for GeoJSON reader

ernstdevreede:
Place a GeoJSON point file in the autowms folder and it will show up.

Querying this service 

https://earthquake.usgs.gov/fdsnws/event/1/query?format=geojson&starttime=2022-07-10&endtime=2022-08-30

and saving the geojson will show:

image


maartenplieger:
Place a GeoJSON point file in the autowms folder and it will show up.

Querying this service, saving the geojson will show:

image

adagucserverEC/CDataReader.cpp
```diff
@@ -1378,7 +1378,7 @@ CDataReader::DimensionType CDataReader::getDimensionType(CDFObject *, CDF::Varia
   }
 
   if (standardName.length() == 0) {
-    CDBDebug("Warning no standard name given for dimension %s, using variable name instead.", variable->name.c_str());
+    // CDBDebug("Warning no standard name given for dimension %s, using variable name instead.", variable->name.c_str());
```
maartenplieger, line 1381: Needed to silence these messages, as the logging became too big.

maartenplieger:
The video looks good! Can you add a functional test to make sure that this feature is tracked with future updates?

PR #229, opened 2022-10-06
Using python:3.9-slim-bullseye as base image

maartenplieger:


lukas-phaf:
Nice that there is a requirements.txt now.
For a next step I would suggest pinning of the Python packages, with controlled semi-automatic updates to the last avaiable versions.

lukas-phaf (approved):

PR #234, opened 2022-11-11
233 support retentionperiod to autoclean files

### LINKED ISSUE #233: Support retentionperiod to autoclean files
Add optional `retentionperiod` and `retentiontype` attributes to FilePath to support auto cleaning of datastreams. Triggered during update of the layer. When a file is added, the cleanup is triggered. The cleanup will start with the oldest files. It will first delete the files from the databasetable, and later from the filesystem. When multiple layers come from the same file, the most strict/shortest retentionperiod is taken.

```xml
<Configuration>
  
  <Layer type="database">
    <FilePath filter=".*\.nc$" retentionperiod="PT5M" retentiontype="filetimedate">/data/adaguc-data/livetimestream/</FilePath>
    
    <Variable>data</Variable>
    <Dimension name="time" interval="PT1M" >time</Dimension>
    <RenderMethod>rgba</RenderMethod>
  </Layer>

</Configuration>
```

maartenplieger:
2.7.11 Rolling archives are now possible by setting the retentionperiod in the FilePath setting of the layer configuration

retentionperiod and retentiontype 
--------------

This feature is available since adaguc-server version 2.7.11

The following dataset configuration configures a dataset which is updated every minute. Without the retentionperiod setting, the archive would grow infinetly large. By setting the retionperiod to `PT1H`, we keep the last hour of data. The files are first removed from the database and then from the filesystem.

- `retentionperiod="PT1H"` means that we want to keep the last hour until now (UTC time). You can configure this according to [iso period](../info/ISO8601.md)
- `retentiontype="filetimedate"` means that the value of the time dimension in the file is read to determine if the file is ready to be deleted.

The retention check is done when a new file is added to the system. 

Manual trigger can be executed with the `--cleanfiles` command:
```
${ADAGUC_PATH}/bin/adagucserver --cleanfiles --config ${ADAGUC_CONFIG},livetimestream
```
Please note that the step above is also done during ingestion of new data.

```xml
<Configuration>
  <Settings enablecleanupsystem="true"/>
  <Layer type="database">
    <FilePath filter=".*\.nc$" retentionperiod="PT1H" retentiontype="filetimedate">/data/adaguc-data/livetimestream/</FilePath>
    
    <Variable>data</Variable>
    <Dimension name="time" interval="PT1M" quantizeperiod="PT1M" quantizemethod="low" >time</Dimension>
    <RenderMethod>rgba</RenderMethod>
  </Layer>

</Configuration>
```

Closes https://github.com/KNMI/adaguc-server/issues/233


CCDFDataModel/CTime.h
```diff
@@ -220,6 +220,11 @@ class CTime {
   static void *currentInitializedVar;
 
   static time_t getEpochTimeFromDateString(CT::string dateString);
+
+  Date subtractPeriodFromDate(CTime::Date date, CT::string period);
```
maartenplieger, line 231: Should add docstrings.

loescornelis:
Is this file meant to be empty? Docker/adaguc-server-addfile.sh

adagucserverEC/CDBFileScannerCleanFiles.cpp
```diff
@@ -6,8 +6,34 @@
...
+      CDBWarning("Unable to create tableName from '%s' '%s' '%s'", dataSource->cfgLayer->FilePath[0]->value.c_str(), dataSource->cfgLayer->FilePath[0]->attr.filter.c_str(), colName.c_str());
+    }
+    CDBDebug("DB: Removing from table %s and dimension %s file %s", tableName.c_str(), colName.c_str(), fileNamestr.c_str());
+    dbAdapter->removeFile(tableName.c_str(), fileNamestr.c_str());
+  }
+}
+
 int CDBFileScanner::cleanFiles(CDataSource *dataSource, int) {
-  CDBDebug("Cleanfiles");
+  if (!dataSource->cfgLayer->FilePath[0]->attr.retentiontype.equals("filetimedate") || dataSource->cfgLayer->FilePath[0]->attr.retentionperiod.empty()) {
```
ernstdevreede: The word filetimedate is not 100% clear. In my opinion it could also mean the timestamp of the file on the file system.
I would suggest something like "datatime" or "timefromdata".
Later we can then use something like "filenametimestamp" when implementing cleaning by ingest date

adagucserverEC/CDBFileScannerCleanFiles.cpp
```diff
@@ -21,14 +47,40 @@ int CDBFileScanner::cleanFiles(CDataSource *dataSource, int) {
...
+
   if (store != NULL && store->getSize() > 0) {
-    for (size_t c = 0; c < store->getColumnModel()->getSize(); c++) {
-      CDBDebug("%d %s", c, store->getColumnModel()->getName(c));
-    }
     for (size_t j = 0; j < store->getSize(); j++) {
       CT::string fileNamestr = store->getRecord(j)->get(0)->c_str();
-      CDBDebug("fileName from DB: %s", fileNamestr.c_str());
+      _removeFileFromTables(fileNamestr, dataSource);
+      int status = remove(fileNamestr.c_str());
```
ernstdevreede: This error message also gets shown when a file with 2 dimensions gets cleaned: the remove for the 2nd dimension fails because the file was already removed.
The resulting error message could cause confusion or might be ignored when there really is a problem.

maartenplieger: A set is used to track if the file is already deleted. Can you check if this works as you expect?

adagucserverEC/CDBFileScannerCleanFiles.cpp
maartenplieger: Changed it to `datatime`

ernstdevreede (approved):

PR #241, opened 2022-12-14
Add support for Mac M1

lukas-phaf:
Made the following changes to support building and running the code on Mac M1:
- CMake changes so that it works with libraries installed through Homebrew.
- Fix C++ code so that it compiles in Clang 14.0.
- Fix C++ code that it works with last version of netcdf.
- Fix a Python issue so that it works with Python 3.9.
- Workaround for race condition in memory leak tracer.
- Fix a bunch of warnings.

Also fix small bug in update dataset script.


cmake/FindPROJ.cmake
```diff
@@ -0,0 +1,45 @@
+# Based on https://github.com/MapServer/MapServer/blob/main/cmake/FindProj.cmake
```
maartenplieger, line 1: Indeed nice to mention that this is based on Mapserver's FindProj.cmake file.

maartenplieger:
I think it is ready to remove the Draft.

maartenplieger:
Hi Lukas,

- Can you change the version to 2.8.0 in the Definition.h file and the Dockerfile?
- Can you add a line to the News.md file saying what 2.8.0 has changed?

ernstdevreede (approved):
Changes work on my Linux environment.
Building succeeds and all tests pass.
Can't test the MAC (neither Intel nor ARM) environment though.

PR #251, opened 2023-02-15
Outlines barbs and text

maartenplieger:


tests/expectedoutputs/TestCSV/test_CSV_windbarbs.gif
lukas-phaf, line 1: - Barbs don't cross middle of circle/station
- I don't think there should be a a wind barb with 5 long barbs, as this is represented by the triangle.

tests/expectedoutputs/TestCSV/test_CSV_windbarbs.png
lukas-phaf, line 1: Shouldn't the triangle always be filled?

tests/expectedoutputs/TestWMS/test_WMSGetMap_KMDS_PointNetCDF_ffdd_windspeed_barb_barb.png
lukas-phaf, line 1: There is a strange raster effect around the shaft. It is quite obvious when you use the "Swipe" view in the github diff.

It looks like the outline is somehow drawn wrong...

adagucserverEC/CImgWarpBilinear.cpp
```diff
@@ -782,32 +801,26 @@ bool IsTextTooClose(std::vector<Point> *textLocations, int x, int y) {
...
 
-  double angle;
-  if (lineY - endY != 0) {
-    angle = atan((-double(lineY - endY)) / (double(lineX - endX)));
+  double angle = atan2(lineX - endX, lineY - endY) - 3.1415 / 2;
+  double angleP = atan2(endY - lineY, endX - lineX) + 3.1415 / 2;
+  float centerX = (lineX + endX) / 2;
+  float centerY = (endY + lineY) / 2;
+
+  if (angle < -3.1415 / 2 || angle > 3.1415 / 2) {
```
lukas-phaf: Define a constant pi and use it everyhwere?
`constexpr double pi = 3.14159265358979323846;`

adagucserverEC/CDrawImage.cpp
```diff
@@ -866,28 +884,29 @@ void CDrawImage::setText(const char *text, size_t length, int x, int y, CColor c
...
 
-void CDrawImage::setTextStroke(const char *text, size_t length, int x, int y, int fgcolor, int bgcolor, int fontSize) {
+void CDrawImage::setTextStroke(int x, int y, float angle, const char *text, const char *fontFile, float fontSize, float strokeWidth, CColor bgcolor, CColor fgcolor) {
   if (currentGraphicsRenderer == CDRAWIMAGERENDERER_CAIRO) {
-    // Not yet supported...
-    setText(text, length, x, y, fgcolor, fontSize);
+    if (bgcolor.a == 0) {
+      drawText(x, y, fontFile, fontSize, angle, text, fgcolor);
+    } else {
+      cairo->drawStrokedText(x, y, -angle, text, fontFile, fontSize * 1.4, strokeWidth, bgcolor, fgcolor);
```
lukas-phaf: Angle is defined different in local drawText versus cairo->drawStrokedText?

adagucserverEC/CCairoPlotter.h
```diff
@@ -123,6 +125,8 @@ class CCairoPlotter {
   void writeToPng32Stream(FILE *fp, unsigned char alpha);
   void writeToWebP32Stream(FILE *fp, unsigned char alpha, int quality);
   void setToSurface(cairo_surface_t *png);
+  void drawStrokedText(int x, int y, double angle, const char *text, const char *fontFile, float fontSize, float strokeWidth, CColor bgcolor, CColor fgcolor);
+  void drawBarb(int x, int y, double direction, double strength, CColor color, CColor outlineColor, bool drawOutline, float lineWidth, bool toKnots, bool flip);
```
lukas-phaf: Move these a couple of lines up to directly below drawText function?

adagucserverEC/CCairoPlotter.cpp
```diff
@@ -653,6 +660,33 @@ void CCairoPlotter::drawText(int x, int y, double angle, const char *text) {
   _drawFreeTypeText(x, y, w, h, angle, text, true);
 }
 
+void CCairoPlotter::drawStrokedText(int x, int y, double angle, const char *text, const char *fontFile, float fontSize, float strokeWidth, CColor bgcolor, CColor fgcolor) {
+  if (fontFile) {
+  }
+  /* https://www.cairographics.org/samples/text/ */
+  cairo_save(cr);
+  cairo_path_t *cp = cairo_copy_path(cr);
```
lukas-phaf: Add a comment to explain this is done to prepend the new elements to the path (or maybe I misunderstood what is going on)...

adagucserverEC/CCairoPlotter.cpp
```diff
@@ -1044,3 +1078,129 @@ void CCairoPlotter::writeToWebP32Stream(FILE *fp, unsigned char, int quality) {
...
+    cairo_set_line_width(cr, 4);
+    cairo_stroke_preserve(cr);
+  }
+
+  // Stroke thin version
+  cairo_set_source_rgba(cr, barbColor.r / 255., barbColor.g / 255., barbColor.b / 255., 1);
+  cairo_set_line_width(cr, 1.5);
+  cairo_stroke(cr);
+
+  // End aend restore
```
lukas-phaf: ??

adagucserverEC/CCairoPlotter.cpp
```diff
@@ -1044,3 +1078,129 @@ void CCairoPlotter::writeToWebP32Stream(FILE *fp, unsigned char, int quality) {
...
+    int nBarbs = (strengthInKnots % 50) / 10;
+    int rest = (strengthInKnots % 50) % 10;
+    int nhalfBarbs;
+    if (rest <= 2) {
+      nhalfBarbs = 0;
+    } else if (rest <= 7) {
+      nhalfBarbs = 1;
+    } else {
+      nhalfBarbs = 0;
+      nBarbs++;
```
lukas-phaf: This is where you can get 5 bars (see my comment on the barb plots).  Same code is in CDrawImage.cpp

adagucserverEC/CCairoPlotter.cpp
```diff
@@ -1044,3 +1078,129 @@ void CCairoPlotter::writeToWebP32Stream(FILE *fp, unsigned char, int quality) {
...
+    }
```
lukas-phaf: A bit scary to touch this code, but this seems to work for the cases around 50 as well, and is simpler:
```
        int strengthInFiveKnots = (strengthInKnots + 2.5) / 5;
        int nPennants2 = strengthInFiveKnots / 10;
        int nBarbs2 = (strengthInFiveKnots % 10) / 2;
        int nhalfBarbs2 = strengthInFiveKnots % 2;
```

This test code gives me some confidence:
```
#include <iostream>

using namespace std;

int main()
{
    for(int i=0; i <=200; i++) {
        int strengthInKnots = i;
        
        // New
        int strengthInFiveKnots = (strengthInKnots + 2.5) / 5;
        int nPennants2 = strengthInFiveKnots / 10;
        int nBarbs2 = (strengthInFiveKnots % 10) / 2;
        int nhalfBarbs2 = strengthInFiveKnots % 2;

//        printf("%d: %d %d %d\n", i, nPennants2, nBarbs2, nhalfBarbs2);

        // Old
        int nPennants = strengthInKnots / 50;
        int nBarbs = (strengthInKnots % 50) / 10;
        int rest = (strengthInKnots % 50) % 10;
        int nhalfBarbs;
        if (rest <= 2) {
          nhalfBarbs = 0;
        } else if (rest <= 7) {
          nhalfBarbs = 1;
        } else {
          nhalfBarbs = 0;
          nBarbs++;
        }
    
//        printf("%d: %d %d %d\n\n", i, nPennants, nBarbs, nhalfBarbs);
        if (nPennants != nPennants2 || nBarbs != nBarbs2 || nhalfBarbs != nhalfBarbs2) {
          printf("%d new : %d %d %d\n", i, nPennants2, nBarbs2, nhalfBarbs2);
          printf("%d old: %d %d %d\n\n", i, nPennants, nBarbs, nhalfBarbs);
            
        }
    }
    return 0;
}
```
        
Output:
```
48 new : 1 0 0
48 old: 0 5 0

49 new : 1 0 0
49 old: 0 5 0

98 new : 2 0 0
98 old: 1 5 0

99 new : 2 0 0
99 old: 1 5 0

148 new : 3 0 0
148 old: 2 5 0

149 new : 3 0 0
149 old: 2 5 0

198 new : 4 0 0
198 old: 3 5 0

199 new : 4 0 0
199 old: 3 5 0
```

adagucserverEC/CCairoPlotter.cpp
```diff
@@ -1044,3 +1078,129 @@ void CCairoPlotter::writeToWebP32Stream(FILE *fp, unsigned char, int quality) {
...
+      double wx4 = wx1 + pos * dx1 / nrPos;
+      double wy4 = wy1 - pos * dy1 / nrPos;
+
+      double ptx[3] = {wx3, hx3, wx4};
+      double pty[3] = {wy3, hy3, wy4};
+
+      cairo_move_to(cr, ptx[0], pty[0]);
+      for (int i = 1; i < 3; i++) {
+        cairo_line_to(cr, ptx[i], pty[i]);
+      }
```
lukas-phaf: Why not use cairo_poly here?

adagucserverEC/CCairoPlotter.cpp
```diff
@@ -1044,3 +1078,129 @@ void CCairoPlotter::writeToWebP32Stream(FILE *fp, unsigned char, int quality) {
...
+    dx1 = cos(direction) * (shaftLength);
+    dy1 = sin(direction) * (shaftLength);
+
+    wx1 = double(x) - dx1;
+    wy1 = double(y) + dy1; // wind barb top (flag side)
+    wx2 = double(x);
+    wy2 = double(y); // wind barb root
+
+    // Draw small center circle
+    cairo_arc(cr, int(wx2), int(wy2), 2, 0, 2 * M_PI);
```
lukas-phaf: Use CairoPlotter::circle() function?

adagucserverEC/CImgWarpBilinear.cpp
lukas-phaf: Also a bit worrying that pi is used with different number of decimal places in the code. This might lead to strange errors...

adagucserverEC/CCairoPlotter.cpp
```diff
@@ -1044,3 +1078,129 @@ void CCairoPlotter::writeToWebP32Stream(FILE *fp, unsigned char, int quality) {
 }
 
 #endif
+
+void CCairoPlotter::drawBarb(int x, int y, double direction, double strength, CColor barbColor, CColor outlineColor, bool drawOutline, float lineWidth, bool toKnots, bool flip) {
```
lukas-phaf: The calculation logic in this function is duplicated in `CDrawImage::_drawBarbGd`, only the drawing itself is different. Assuming we need to keep both, what is the best way to reuse the logic?

We could abstract out the draw logic, but I think that will get messy.

What about separating out the math instead, into a BarbCalculation class, which does only math.
So based on wind strength and barb x,y position it can calculate the number of penants, coordinates of these pennants, etc.

The main code would still have the loops over for example the nBarbs, but it would have only one line to get the coordinates, the rest would be drawing code.

adagucserverEC/CDrawImage.h
```diff
@@ -207,7 +210,7 @@ class CDrawImage {
   void setDisc(int x, int y, float discRadius, CColor fillColor, CColor lineColor);
   void setEllipse(int x, int y, float discRadiusX, float discRadiusY, float rotation, CColor fillColor, CColor lineColor);
   void setTextDisc(int x, int y, int discRadius, const char *text, const char *fontfile, float fontsize, CColor textcolor, CColor fillcolor, CColor lineColor);
-  void setTextStroke(const char *text, size_t length, int x, int y, int fgcolor, int bgcolor, int fontSize);
+  void setTextStroke(int x, int y, float angle, const char *text, const char *fontFile, float fontsize, float strokeWidth, CColor bgcolor, CColor fgcolor);
```
lukas-phaf, line 211: It feels strange that a function with this name actually draws text.

data/config/datasets/adaguc.testCSVReader.xml
```diff
@@ -66,6 +66,16 @@
...
   </Layer>
+
+  <Layer type="database">
+    <FilePath filter="^.*\.csv$">{ADAGUC_PATH}/data/datasets/csv_wind/csvexample_windbarbs.csv</FilePath>
+    <Name>windallspeeds</Name>
+    <Title>windallspeeds</Title>
+    <Variable>ff</Variable>
+    <Variable>dd</Variable>
+    <Styles>windbarb,windvector</Styles>
+  </Layer>
```
lukas-phaf, line 78: A bit strange to have this test in testCSVReader, but it probably beats having a whole new test for it...

adagucserverEC/CImgWarpBilinear.cpp
```diff
@@ -741,18 +747,31 @@ int CImgWarpBilinear::set(const char *pszSettings) {
               kvp[1].setSize(7);
               textcolor = CColor(kvp[1].c_str());
             }
+            if (kvp[0].equals("textstrokecolor")) {
+              kvp[1].setSize(7);
+              textstrokecolor = CColor(kvp[1].c_str());
+            }
```
lukas-phaf, line 750: Can you explain how this linesetting bit works (probably easiest in a call or in person...)

adagucserverEC/CImgWarpBilinear.cpp
lukas-phaf: There are two cases to avoid upside-down text?

adagucserverEC/CImageDataWriter.cpp
```diff
@@ -1306,7 +1306,7 @@ int CImageDataWriter::createAnimation() {
   return 0;
 }
 
-void CImageDataWriter::setDate(const char *szTemp) { drawImage.setTextStroke(szTemp, strlen(szTemp), drawImage.Geo->dWidth - 170, 5, 240, 254, 0); }
+void CImageDataWriter::setDate(const char *szTemp) { drawImage.setTextStroke(drawImage.Geo->dWidth - 170, 5, 0, szTemp, NULL, 12.0, 0.75, CColor(0, 0, 0, 255), CColor(255, 255, 255, 255)); }
```
lukas-phaf, line 1305: Strange name for this function?

adagucserverEC/CImageDataWriter.cpp
```diff
@@ -1599,12 +1599,25 @@ int CImageDataWriter::warpImage(CDataSource *dataSource, CDrawImage *drawImage)
...
+              if (contourLine->attr.textsize.empty() == false) {
+                bilinearSettings.printconcat("textsize(%s)$", contourLine->attr.textsize.c_str());
+              }
+              if (contourLine->attr.textstrokewidth.empty() == false) {
+                bilinearSettings.printconcat("textstrokewidth(%s)$", contourLine->attr.textstrokewidth.c_str());
+              }
+
+              if (contourLine->attr.textstrokecolor.empty() == false) {
+                bilinearSettings.printconcat("textstrokecolor(%s)$", contourLine->attr.textstrokecolor.c_str());
+              }
```
lukas-phaf, line 1607: Why is this communicated by putting it into a single string, and later parsing it out again?

tests/expectedoutputs/TestCSV/test_CSV_windbarbs.png
maartenplieger: The triangle is now filled

tests/expectedoutputs/TestCSV/test_CSV_windbarbs.gif
maartenplieger: Indeed, now fixed.

adagucserverEC/CCairoPlotter.cpp
maartenplieger: This ensures that the same path settings, like line thickness and colors are re-used.

adagucserverEC/CCairoPlotter.cpp
maartenplieger: Same, This ensures that the same path settings, like line thickness and colors are re-used.

adagucserverEC/CCairoPlotter.cpp
maartenplieger: You are right. I think we would need remove GD and the drawing logic from CDrawImage in a next step.

adagucserverEC/CDrawImage.h
maartenplieger: What would you suggest?

maartenplieger: drawStrokedText?

data/config/datasets/adaguc.testCSVReader.xml
maartenplieger: Yes, but I would keep it here. The CSV reader is also a showcase on how to plot wind.

adagucserverEC/CImageDataWriter.cpp
maartenplieger: That is legacy. Now it would be better to directly pass the object. Would say this is technical depth.

adagucserverEC/CImgWarpBilinear.cpp
maartenplieger: At least PI is now defined at a single place :+1:

adagucserverEC/CDrawImage.cpp
maartenplieger: Yes, drawText uses freetype and drawstroked text uses cairo...

adagucserverEC/CCairoPlotter.h
maartenplieger: Done

adagucserverEC/CCairoPlotter.cpp
maartenplieger: Removed

adagucserverEC/CCairoPlotter.cpp
maartenplieger: Fixed

adagucserverEC/CImgWarpBilinear.cpp
maartenplieger: Yes, I can explain.

tests/expectedoutputs/TestCSV/test_CSV_windbarbs.gif
lukas-phaf: I fixed the barbs issue in the Gd version. The "middle" of circle is till there, but do we care?

adagucserverEC/CImgWarpBilinear.cpp
lukas-phaf: Need to take another look. There is another pi in:
CCDFPNGIO.cpp

Another issue is dat <math.h> also defines M_PI (at least on my Mac, I think this is not C++11 compliant), but with different precision than your M_PI. It depends on which includes you have which one you get.

A bit of a mess. Shall we rely on the math.h version (assuming it works across platforms)?

tests/expectedoutputs/TestCSV/test_CSV_windbarbs.gif
maartenplieger: Looks good to me.

adagucserverEC/CImgWarpBilinear.cpp
maartenplieger: Yes that would be best.

adagucserverEC/CImgWarpBilinear.cpp
maartenplieger: This ensures no garbage is going into CColor constructor.

adagucserverEC/CImgWarpBilinear.cpp
lukas-phaf: See https://github.com/KNMI/adaguc-server/pull/251/commits/90d5e83de0564015e16b7080ded62c5b3a9efad1

adagucserverEC/CDrawImage.h
lukas-phaf: Let's leave it, as this seems to be the naming convention in CDrawImage.

adagucserverEC/CDataPostProcessor.cpp
```diff
@@ -1067,7 +1067,7 @@ int CDPPBeaufort::execute(CServerConfig::XMLE_DataPostProc *proc, CDataSource *d
 /************************/
 const char *CDPPToKnots::className = "CDPPToToKnots";
 
-const char *CDPPToKnots::getId() { return "toknots"; }
+const char *CDPPToKnots::getId() { return ""; }
```
lukas-phaf: Was this intended?

adagucserverEC/CDrawImage.cpp
```diff
@@ -466,26 +446,39 @@ void CDrawImage::drawBarb(int x, int y, double direction, double strength, CColo
...
   }
 
   line(wx1, wy1, wx2, wy2, lineWidth, color);
 }
 
+void CDrawImage::drawBarb(int x, int y, double direction, double strength, CColor color, float lineWidth, bool toKnots, bool flip, bool drawText) {
+  if (currentGraphicsRenderer == CDRAWIMAGERENDERER_GD) {
+    _drawBarbGd(x, y, direction, strength, color, lineWidth, toKnots, flip);
+    if (drawText) {
+      // TODO: DRAW TEXT
```
lukas-phaf, line 473: We can leave this TODO for now?

tests/expectedoutputs/TestCSV/test_CSV_timesupport2018-12-04T12:00:00Z.png
lukas-phaf, line 1: Is it correct that the windspeed shows up as a number here?

adagucserverEC/CDataPostProcessor.cpp
maartenplieger: Ouch that is a mistake. Should be toknots...

adagucserverEC/CDrawImage.cpp
maartenplieger: Yes

tests/expectedoutputs/TestCSV/test_CSV_timesupport2018-12-04T12:00:00Z.png
maartenplieger: Jep that is good.

lukas-phaf (approved):

PR #254, opened 2023-03-15
Add UoM field to DescribeCoverage output

lukas-phaf:
This feature was requested by KDP, for use in the EDR API.


maartenplieger (approved):
Looking good. Thanks for adding.

PR #259, opened 2023-05-11
Adding Strict-Transport-Security header when running over https

maartenplieger:


python/python_fastapi_server/main.py
```diff
@@ -30,6 +31,17 @@ async def add_process_time_header(request: Request, call_next):
     return response
 
 
+@app.middleware("http")
+async def add_hsts_header(request: Request, call_next):
+    response = await call_next(request)
+    if "EXTERNALADDRESS" in os.environ:
+        external_address = os.environ["EXTERNALADDRESS"]
```
ernstdevreede, line 38: There might be cases where a server is running under http and https. After automatically setting this header once for an https: request that browser will (for the next year) not be able to reach that same server over http:

If I understand well.

So it might be good to set this option (or not set it) depending on a runtime environment variable.

ernstdevreede (approved):
I'm going to approve this.

PR #270, opened 2023-06-15
Fastapi maps

ernstdevreede:
This is a rudimentary version of EDR code for ADAGUC's python server.
This code provides all required API calls, but provides only the "position" query (for point timeseries).
It should work on model data and observed gridded data (radar for example).

EDR configuration is configured by adding a few lines to the dataset configuration files.
A few config files are attached to this MR, to show how this could work.

The code currently does not very well handle the case where there is data for multiple reference times online. This still needs to be fixed.

You can test the API either with the MetOffice's EDR tester (https://labs.metoffice.gov.uk/edr/static/html/query.html) or the internal SWAGGER UI (http://localhost:8080/docs) or even with GeoWeb's time series code (needs some configuring).

See [edr_configs.zip](https://github.com/KNMI/adaguc-server/files/11757690/edr_configs.zip)

Install these configs, ingest some data (HARM_N25, RADAR) and it might work.

There is also a rudimentary test for the EDR stuff, which also shows how to configure EDR.


python/edr_package/edr_pydantic_classes.egg-info/PKG-INFO
```diff
@@ -0,0 +1,13 @@
...
+Summary: The Pydantic models for EDR datatypes
+Home-page: https://gitlab.com/KNMI/data-platform/observation-api/internal-data-api
+Author: KNMI Data Platform
+Author-email: dataplatform@knmi.nl
+License: MIT
+Platform: UNKNOWN
+Requires-Python: >=3.8.0
+
+UNKNOWN
+
```
lukas-phaf, line 13: I think the files in the edr_pydantic_classes.egg-info directory are automatically generated when you install the package, so they should not be in git.

python/edr_package/setup.py
```diff
@@ -0,0 +1,34 @@
...
+if env_suffix:
+    NAME += f"-{env_suffix}"
+logger.debug(f"Package name: {NAME}")
+
+setuptools.setup(
+    name=NAME,
+    version="0.0.9",
+    author="KNMI Data Platform",
+    author_email="dataplatform@knmi.nl",
+    url="https://gitlab.com/KNMI/data-platform/observation-api/internal-data-api",
```
lukas-phaf, line 26: It is probably not a good idea to leave this in.

python/python_fastapi_server/routers/edr_types/extent.py
```diff
@@ -0,0 +1,39 @@
...
+class Temporal(BaseModel):
+    interval: list[str]
+    trs: str
+
+    @classmethod
+    def custom_init(cls, interval: list[str], trs: str):
+        return cls(interval=interval, trs=trs)
+
+
+class Extent(BaseModel):
```
lukas-phaf, line 22: There is a lot of duplcate (triplicate?) types now. For example class Extent (and it member Spatial and Temporal) is defined in `edr_pydantic_classes` package, in `edr_types` directory, and in the file `ogcapifeatures_1_model.py`. 

They are all slightly different versions, but my guess is only one is neeeded?

python/python_fastapi_server/routers/edr.py
```diff
@@ -0,0 +1,920 @@
...
+                "time_interval":
+                edr_collection.attrib.get("time_interval"),
+                "z_interval":
+                edr_collection.attrib.get("z_interval"),
+                "parameters": [
+                    inst.strip()
+                    for inst in edr_collection.text.strip().split(",")
+                ],
+            }
+    return edr_collections
```
lukas-phaf, line 155: Drop this?

python/python_fastapi_server/routers/edr.py
```diff
@@ -0,0 +1,920 @@
...
+                        "z_interval":
+                        edr_collection.attrib.get("z_interval"),
+                        "parameters": [
+                            inst.strip()
+                            for inst in edr_collection.text.strip().split(",")
+                        ],
+                    }
+        except ParseError:
+            pass
+    return edr_collections
```
lukas-phaf, line 126: It feels wrong to get this from the XML files directly. Should this data not be requested from adaguc server?

python/python_fastapi_server/routers/edr.py
```diff
@@ -0,0 +1,920 @@
...
+        coords: str,
+        datetime: Optional[str] = None,
+        parameter_name: str = Query(alias="parameter-name"),
+):
+    allowed_params = ["coords", "datatime", "parameter-name"]
+    extra_params = [k for k in request.query_params if k not in allowed_params]
+    extra_dims = ""
+    if len(extra_params):
+        for extra_param in extra_params:
+            extra_dims += f"&DIM_{extra_param}={request.query_params[extra_param]}"
```
lukas-phaf, line 206: What are the extra parameters used for exactly? Do you have an example?

python/python_fastapi_server/routers/edr.py
```diff
@@ -0,0 +1,920 @@
...
+    edr_collections = get_edr_collections()
+    dataset = edr_collections[collection_name]["dataset"]
```
lukas-phaf, line 208: This scans all XML files for every query, correct?

python/python_fastapi_server/routers/edr.py
```diff
@@ -0,0 +1,920 @@
...
+        dataset,
+        None,
+        [coord["lon"], coord["lat"]],
+        parameter_name,
+        datetime,
+        extra_dims,
+    )
+    if resp:
+        dat = json.loads(resp)
+        return GeoJSONResponse(covjson_from_resp(dat))
```
lukas-phaf, line 226: Tag the CoverageJSON response as `application/prs.coverage+json` instead of as geojson?

lukas-phaf, line 226: If you use the Pydantic 2 version of the covjson-pydantic (soon to be merged), you can just do `return covjson_from_resp(dat)` and it will work fine. The explicit `.json()` call is no longer needed for performance.

You will need to add `response_model_exclude_none=True,` to the endpoint parameters.

python/python_fastapi_server/routers/edr.py
```diff
@@ -0,0 +1,920 @@
...
+        [coord["lon"], coord["lat"]],
+        parameter_name,
+        datetime,
+        z,
+    )
+    if resp:
+        dat = json.loads(resp)
+        return GeoJSONResponse(covjson_from_resp(dat))
+
+    raise EdrException(code=400, description="No data")
```
lukas-phaf, line 268: Re-use non-instance version?

python/python_fastapi_server/routers/edr.py
```diff
@@ -0,0 +1,920 @@
...
+                    param["explode"] = False
+                    param["schema"] = {
+                        "type": "string",
+                    }
+    if "CompactAxis" in api["components"]["schemas"]:
+        comp = api["components"]["schemas"]["CompactAxis"]
+        if "exclusiveMinimum" in comp["properties"]["num"]:
+            comp["properties"]["num"]["exclusiveMinimum"] = False
+
+    return api
```
lukas-phaf, line 728: Is this a hack to make it pass certification?

python/python_fastapi_server/main.py
```diff
@@ -8,15 +12,15 @@
...
 
-from configure_logging import configure_logging
 from routers.autowms import autowms_router
 from routers.healthcheck import health_check_router
 from routers.middleware import FixSchemeMiddleware
 from routers.ogcapi import ogcApiApp
 from routers.opendap import opendapRouter
 from routers.wmswcs import testadaguc, wmsWcsRouter
+from routers.edr import edrApiApp
+from routers.maps import create_maps_routes
```
lukas-phaf, line 22: Is this for the OGC Maps API?

python/python_fastapi_server/routers/edr.py
ernstdevreede: Dropped

python/python_fastapi_server/routers/edr.py
ernstdevreede: Done

python/python_fastapi_server/routers/edr_types/extent.py
ernstdevreede: There are 3 OGCAPI packages: OGCAPI Features, OGCAPI Maps and OGCAPI-EDR.  All have their own set of types, at the moment.
Confusingly the OGCAPI-Maps code uses the edr_types set. This should be fixed in the fastapi-maps-only branch.

Eventually some type of type consolidation could be considered between the 3 OGCAPI api's.

python/python_fastapi_server/routers/edr.py
ernstdevreede: At this moment it seems better to get this collection-wide set of configurations from the XML files, so they are equal for all parameters in an EDR collection (an EDR collection being a subset of an ADAGUC dataset with the same set of dimensions)

python/python_fastapi_server/routers/edr.py
ernstdevreede: They are custom dimensions or the data. Because they are custom, they can't be coded into the method's parameter list by name, they have to be extracted from the query parameter.

I renamed them to custom_dims

PR #275, opened 2023-08-09
Improve Python requirements handling.

lukas-phaf:
We are currently not pinnig the versions (even the major versions) of Python packages. This means that package releases can easily break Adaguc.

To avoid this, I changed the handling of requirements to use pip-tools:
https://github.com/jazzband/pip-tools
It is installed using:
`pip install pip-tools`

The top level requirements (dependencies of these requirements don't need to be specified) are now given in a `requirements.in` file, using the compatibility operator (see https://peps.python.org/pep-0440/#compatible-release).
This allows semi-automatic updates of minor version (which are supposed to be backward compatible), but now major versions.

The actual `requirements.txt` was automatically generated using `pip-compile` (the exact command is given in the `requirements.in` file), pinning the packages to specific versions.

Both `requirements.in` and `requirements.txt` files are committed to git. When you think updating the dependencies is appropriate, you run the `pip-compile` command, test, and push the resulting `requirements.txt` file.

A similar change was made for the `requirements-dev.txt` file. 

Note that the `requirements.in` and `requirements.txt` files are comparable to the `package.json` and `package-lock.json` files in `npm`.

Another useful command in pip-tools is `pip-sync`: it makes sure the packages in `requirements.txt` file are installed in your Python environment, and nothing else. This is different from `pip install`, which just keeps adding packages to the existing environment.


maartenplieger (approved):
Looks good!

PR #278, opened 2023-08-21
Prevent double logging

maartenplieger:
Fixed two issues since we are using fast-api:
- Logging showed up twice
- Process had no timeout

### DIFF
--- a/python/lib/adaguc/CGIRunner.py
+++ b/python/lib/adaguc/CGIRunner.py
@@ -42,6 +42,7 @@ class CGIRunner:
             (processOutput, processError) = process.communicate(timeout=timeout)
         except TimeoutExpired:
             process.kill()
+            process.communicate()
             output.write(b'TimeOut')
             return 1, [], None
 
--- a/python/lib/adaguc/CGIRunner.py
+++ b/python/lib/adaguc/CGIRunner.py
@@ -4,7 +4,7 @@
 """
 
 import sys
-from subprocess import PIPE, Popen, STDOUT
+from subprocess import PIPE, Popen, STDOUT, TimeoutExpired
 from threading import Thread
 import os
 import io
@@ -22,7 +22,7 @@ class CGIRunner:
       Run the CGI script with specified URL and environment. Stdout is captured and put in a BytesIO object provided in output
     """
 
-    def run(self, cmds, url, output, env=[],  path=None, isCGI=True):
+    def run(self, cmds, url, output, env=[],  path=None, isCGI=True, timeout=300):
         localenv = {}
         if url != None:
             localenv['QUERY_STRING'] = url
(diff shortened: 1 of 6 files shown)


belentorrente:
Looks good to me.

python/lib/adaguc/CGIRunner.py
```diff
@@ -37,8 +37,14 @@ def run(self, cmds, url, output, env=[],  path=None, isCGI=True):
         # Execute adaguc-server binary
         ON_POSIX = 'posix' in sys.builtin_module_names
         process = Popen(cmds, stdout=PIPE, stderr=PIPE, env=localenv, close_fds=ON_POSIX)
-        (processOutput, processError) = process.communicate()
- 
+        
+        try:
+            (processOutput, processError) = process.communicate(timeout=timeout)
+        except TimeoutExpired:
+            process.kill()
```
lukas-phaf, line 44: According to [this](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate), you also need to do a `proc.communicate()` to do a proper cleanup?

python/python_fastapi_server/configure_logging.py
```diff
@@ -3,7 +3,17 @@
 """Configures logging for the adaguc-server python wrapper"""
 import sys
 
+def run_once(f):
+    def wrapper(*args, **kwargs):
+        if not wrapper.has_run:
+            wrapper.has_run = True
```
lukas-phaf: Change the order like [here](https://gist.github.com/jasongrout/3804691).
If `configure_logging` somehow fails, it will not run again, as `has_run` is set to True before it is run.

python/python_fastapi_server/routers/wmswcs.py
```diff
@@ -46,6 +46,7 @@ async def handle_wms(req: Request, ):
     # Run adaguc-server
     status, data, headers = adaguc_instance.runADAGUCServer(query_string,
                                                             env=adagucenv,
+                                                            showLogOnError=False,
```
lukas-phaf, line 49: Was this change intended?

python/python_fastapi_server/configure_logging.py
```diff
@@ -3,7 +3,17 @@
...
+            return f(*args, **kwargs)
 
+    wrapper.has_run = False
+    return wrapper
+
+
+@run_once
 def configure_logging(logging):
```
lukas-phaf, line 6: Not clear to me why it is called twice. The solution is a bit... brutal.

python/python_fastapi_server/routers/wmswcs.py
maartenplieger: Yes, the logging is in wmswcs.py always displayed, see line 57. It does not need to show it once more when there is an error.

python/lib/adaguc/CGIRunner.py
maartenplieger: yes, good one. Clean kill.

python/python_fastapi_server/configure_logging.py
maartenplieger: Okay, I think I added the @wraps correctly.

python/python_fastapi_server/configure_logging.py
maartenplieger: I found that gunicorn or uvicorn also add a logging handler.

I tried to add logging.getLogger().handlers.clear() in configure_logging.

This looks like a better solution to me?

python/python_fastapi_server/configure_logging.py
maartenplieger: Now completely removed the run_once part.

lukas-phaf (approved):

PR #288, opened 2023-09-06
Optimize addfile script and clean based on reference_time for models for WIK

maartenplieger:
**Version 2.13.0 2023-09-20**
- The script adaguc-server-addfile.sh is now able to find for a given file the corresponding dataset via the commandline option autofinddataset
- Cleaning of model data is now done based on the forecast_reference_time dimension
- Logging can be made less verbose by adding --verboseoff

# How test:

- copy the file `data/datasets/testdata.nc` to the `/data/adaguc-data/ folder`
- Make a dataset config in the /data/adaguc-datasets folder like:

```xml
<Configuration>
  <Layer type="database">
    <FilePath filter="">/data/adaguc-data/testdata.nc</FilePath>
     <Variable>testdata</Variable>
    <Styles>auto</Styles>
  </Layer>
</Configuration>
``` 
- Make sure your adaguc environment is correctly setup
- run the command `bash ./Docker/adaguc-server-addfile.sh -f /data/adaguc-data/testdata.nc`

It should auto detect to which dataset this file belongs.


adagucserverEC/adagucserver.cpp
```diff
@@ -204,15 +290,35 @@ int _main(int argc, char **argv, char **) {
...
-    if (status != 0) {
-      CDBError("Error occured in updating the database");
+    std::set<std::string> datasetsToScan;
+
+    if (automaticallyFindMatchingDataset) {
+      datasetsToScan = findDataSetsToScan(layerPathToScan, verbose);
+      CDBDebug("Datasets to scan: %d", datasetsToScan.size());
+    } else {
+      // Empty string to make for loop work.
+      datasetsToScan.insert("");
```
ernstdevreede, line 305: This pushes an empty string to the set.
Is an empty string as dataset name handled correctly  in the for loop  (lines 307-321)?

maartenplieger: Yes that will work. Set `CRequestConfigFromEnvironment(CRequest *request, CT::string dataset = "")`; will work just as before when the dataset is empty.

ernstdevreede: OK

ernstdevreede (approved):

PR #324, opened 2024-01-08
WIP: Coverage for multidimentional data

belentorrente:
As I wrote in the chat last month: Just some thoughts on the DescribeCoverage problem when it comes to including multidimensional layers. I thought maybe adding multiple AxisDescriptions would be the way to go, based on something I found in the documentation: `The optional and repeatable axisDescription/AxisDescription element (...) describes an additional parameter (that is, an independent variable besides space and time), and the valid values of this parameter, which GetCoverage requests can use to select subsets of a coverage offering`.  However, there is some info about the layer dimensions that it not straightforward to get (for example, min and max for a given dimension), so I am a bit stuck there. Or maybe I could make a different request to provide the metadata that we do have, independently of this standard. Another thing I've observed is that one AxisDescription for the bands of a raster is hardcoded everywhere. Typing this here so I don't forget, we can discuss later.

### DIFF
--- a/adagucserverEC/CXMLGen.cpp
+++ b/adagucserverEC/CXMLGen.cpp
@@ -1492,6 +1492,43 @@ int CXMLGen::getWCS_1_0_0_Capabilities(CT::string *XMLDoc, std::vector<WMSLayer 
   return 0;
 }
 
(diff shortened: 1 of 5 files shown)


belentorrente:
Superseded by this other PR https://github.com/KNMI/adaguc-server/pull/332

PR #332, opened 2024-01-31
Draft: WCS Coverage for multidimensional layers

belentorrente:
As I wrote in the chat last month: Just some thoughts on the DescribeCoverage problem when it comes to including multidimensional layers. I thought maybe adding multiple AxisDescriptions would be the way to go, based on something I found in the documentation: `The optional and repeatable axisDescription/AxisDescription element (...) describes an additional parameter (that is, an independent variable besides space and time), and the valid values of this parameter, which GetCoverage requests can use to select subsets of a coverage offering`.  However, there is some info about the layer dimensions that it not straightforward to get (for example, min and max for a given dimension), so I am a bit stuck there. Or maybe I could make a different request to provide the metadata that we do have, independently of this standard. Another thing I've observed is that one AxisDescription for the bands of a raster is hardcoded everywhere. Typing this here so I don't forget, we can discuss later.


tests/expectedoutputs/TestWCS/test_WCSDescribeCoverage_Multidimdata.xml
```diff
@@ -0,0 +1,179 @@
...
+      <RangeSet>
+        <name>dimensions</name>
+        <label>dimensions</label>
+        <axisDescription>
+          <AxisDescription>
+            <name>"member"</name>
+            <label>"member"</label>
+            <values>
+              <interval>
+                <min>"member1"</min>
```
maartenplieger: Would it be possible to list all values for these dimensions?

adagucserverEC/CXMLGen.cpp
```diff
@@ -1492,6 +1522,59 @@ int CXMLGen::getWCS_1_0_0_Capabilities(CT::string *XMLDoc, std::vector<WMSLayer
...
+  coverage offering.
+  */
+  XMLDoc->concat("    <rangeSet>\n"
+                 "      <RangeSet>\n"
+                 "        <name>dimensions</name>\n"
+                 "        <label>dimensions</label>\n");
+  // Dims
+  for (size_t d = 0; d < layer->dimList.size(); d++) {
+    WMSLayer::Dim *dim = layer->dimList[d];
+    if (dim->name.indexOf("time") != -1) {
```
maartenplieger: Would it be possible to show for time the min/max and resolution?

When I do describecoverage on a KNMI Radar file there is no time dimension in the describecoverage. I think we have to accomodate for a webclient, for a webclient it would be useful to have this information to generate a UI.

tests/expectedoutputs/TestWCS/test_WCSDescribeCoverage_Multidimdata.xml
belentorrente: I think this is solved now

adagucserverEC/CXMLGen.cpp
belentorrente: The resolution in the same way that we show in the getCapabilities description?

maartenplieger: yes!

belentorrente:
Superseded by another PR

PR #344, opened 2024-03-04
Optimize bbox

maartenplieger:


belentorrente (approved):

adagucserverEC/CImageWarper.cpp
```diff
@@ -23,57 +23,15 @@
  *
  ******************************************************************************/
 
+#include "./Types/ProjectionKey.h"
+#include "./Types/ProjectionStore.h"
```
lukas-phaf, line 27: Is the `./` really needed? I thought `#include` with quotes are always checked relative from the current file?

adagucserverEC/CXMLGen.cpp
```diff
@@ -267,13 +267,21 @@ int CXMLGen::getProjectionInformationForLayer(WMSLayer *myWMSLayer) {
...
 
   CGeoParams geo;
-  CImageWarper warper;
+
   int status;
   for (size_t p = 0; p < srvParam->cfg->Projection.size(); p++) {
     geo.CRS.copy(srvParam->cfg->Projection[p]->attr.id.c_str());
 
+#ifdef MEASURETIME
+    StopWatch_Stop("start initreproj %s", geo.CRS.c_str());
```
lukas-phaf, line 276: Maybe I don't understand the `StopWatch` convention, but shouldn't this be `Start`?

Can you check all the `StopWatch` statements you added?

adagucserverEC/CXMLGen.cpp
```diff
@@ -937,7 +957,7 @@ int CXMLGen::getWMS_1_1_1_Capabilities(CT::string *XMLDoc, std::vector<WMSLayer
 
       for (size_t lnr = 0; lnr < myWMSLayerList->size(); lnr++) {
         WMSLayer *layer = (*myWMSLayerList)[lnr];
-        if (layer->group.equals(groupKeys[groupIndex].c_str())) {
+        if (layer->group.equals(&groupKeys[groupIndex])) {
```
lukas-phaf, line 960: Why not define an equals with a `const` reference and use that, instead of adding more pointers variants? Unless these are used somewhere else as well?

hclasses/CTString.cpp
```diff
@@ -526,6 +526,16 @@ namespace CT {
     return equals(_string.useStack ? _string.stackValue : _string.heapValue, _string.privatelength);
   }
 
+  bool string::equals(std::string _string) const {
+    if (allocated == 0) return false;
+    return equals(_string.c_str(), _string.length());
+  }
+
+  bool string::equals(std::string *_string) const {
```
lukas-phaf, line 534: Define function with (const) reference instead?

adagucserverEC/CImageWarper.cpp
```diff
@@ -450,11 +390,12 @@ int CImageWarper::_findExtentSynchronized(CDataSource *dataSource, double *dfBBO
 
         double testPosX = dfBBOX[0] * (1 - stepX) + dfBBOX[2] * stepX;
         double testPosY = dfBBOX[1] * (1 - stepY) + dfBBOX[3] * stepY;
+        bool projError = false;
         double inY = testPosY;
         double inX = testPosX;
-        bool projError = false;
 
         if (testPosX == testPosX && testPosY == testPosY) {
```
lukas-phaf, line 397: ??

adagucserverEC/Types/ProjectionStore.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+ProjectionStore *ProjectionStore::getProjectionStore() { return &projectionStore; }
+
+ProjectionStore::ProjectionStore() {}
+ProjectionStore::~ProjectionStore() { clear(); }
+
+pthread_mutex_t ProjectionKey_clear;
+void ProjectionStore::clear() { keys.clear(); }
+
+int ProjectionStore::findExtentForKey(ProjectionKey pKey, BBOX *bbox) {
```
lukas-phaf, line 43: Why not define a less than operator on pKey, so you can just stuff them in a map?

adagucserverEC/CImageWarper.h
```diff
@@ -72,10 +51,11 @@ class CImageWarper {
   //     int _decodeCRS(CT::string *CRS);
   std::vector<CServerConfig::XMLE_Projection *> *prj;
   bool initialized;
-  int _findExtentSynchronized(CDataSource *dataSource, double *dfBBOX);
+
   int _initreprojSynchronized(const char *projString, CGeoParams *GeoDest, std::vector<CServerConfig::XMLE_Projection *> *_prj);
 
 public:
+  int findExtentUnSynchronized(CDataSource *dataSource, double *dfBBOX);
```
lukas-phaf, line 58: Why now public?

adagucserverEC/CImageWarper.cpp
lukas-phaf: Add comment that this is NaN check.

adagucserverEC/Types/ProjectionStore.cpp
lukas-phaf: Info:
https://cplusplus.com/forum/general/145841/
https://en.cppreference.com/w/cpp/utility/tuple/make_tuple

Issue #282, opened 2023-08-22
WMS HTTP Status codes that provide better feedback to the users.


Jeffrey-Vervoort-KNMI:
Although the WMS is before REST standards, it would be nice to have similar HTTP codes for monitoring and as feedback to the users of the service.

## Expected Behavior
When a user gives invalid input the server responds with a HTTP status code 4xx and a message.

## Current Behavior
When a user gives invalid input the server responds with a clear message about what is wrong with the input, but a different status code than I would expect (200/500). 
For example, when giving `temperatured` as input to the `COVERAGE` parameter. The status code is `500`. And the message is `Layer [temperatured] not found WCS GetCoverage Request failed`. Other examples are in the table below.

Parameter | Input | Status code | Expected
-- | -- | -- | --
COVERAGE | temperatured | 500 | 404
dataset | Tx1d | 500 | 404
BBOX | 3.0,50.0,7.0,99999999999999.0 | 500 | 422
Time | 2022-sdfsd-22T00:00:00Z/2022-07-23T00:00:00Z | 500 | 422
CRS | EPSG:4326d | 500 | 422
RESX/RESY | 0.000004 | 500 | 422
RESX/RESY | 0.d4 | 200 | 422
SERVICE | Tx1d | 200 | 422
REQUEST | GetCoveraged | 200 | 422

Jeffrey-Vervoort-KNMI:
What I came across but did not see as part of the issue, but still might be interesting to know is:

- BBOX outputs NetCDF on incorrect value, e.g. “3.0,5sdfew0.0,7.0,54.0”
- FORMAT always outputs NetCDF on incorrect value, e.g. "arandomstring" as input
- TIME almost always outputs a NetCDF on very strange datetimes: 2022-027-22T00:00:00Z/2022-07-23T00:00:00Z
- ELEVATION always outputs a NetCDF also on random input: "3289urjf"

maartenplieger:
Fixedi n Version 2.24.0 2024-07-09

PR #347, opened 2024-03-05
Introduce 404 and 422 status codes for wrong requests

belentorrente:
This MR tackles some of the issues mentioned here: https://github.com/KNMI/adaguc-server/issues/282

In particular, the correct status codes have been implemented in the following cases:

- Wrong coverage (404 instead of 500)
- Wrong dataset (404 instead of 500)
- Wrong service (422 instead of 200)
- Wrong request (422 instead of 200)

All other cases (BBOX, Time, CRS, RESX/RESY) are currently being ignored when incorrect and will be tackled separately in the future.


ernstdevreede (dismissed):

adagucserverEC/Definitions.h
```diff
@@ -188,3 +188,10 @@
 
 // #define ADAGUC_USE_GDAL
 #endif
+
+// HTTP status codes
+// Success
+#define SUCCESS_200_OK 0
```
maartenplieger: Would it be an idea to prefix these three with  HTTP_STATUSCODE_<...> ?

belentorrente: Sure, will change that!

maartenplieger (approved):

PR #348, opened 2024-03-06
Add PGBouncer

mgrunbauer:
Connecting to the Postgres database takes time. Running Adaguc locally in Docker, connecting takes anywhere between 3 and 20 ms. Running Adaguc in AWS, connecting can take between 20-100 ms.
This (small) overhead happens during every request made to Adaguc, because adaguc-server is a "one-shot" application and does not keep a connection pool open.

This MR adds:
- [pgbouncer](https://www.pgbouncer.org/), which connects to the Postgres server and keeps a connection pool available. This runs in the same docker container as Adaguc server. Adaguc server can then connect to pgbouncer on `localhost`.
- [supervisord](http://supervisord.org/), which can run both adaguc-server and pgbouncer.
- A new environment variable called `PGBOUNCER_ENABLE`. When set to `true|TRUE`, supervisord will launch both adaguc-server and pgbouncer. If the environment variable is omitted (or not set to `true`), supervisord will only launch adaguc-server. The first log line of the adaguc-container will show what it will run.

Using this setup, connection time drops to >1-3 ms even when running on AWS.


Docker/run_supervisord.sh
```diff
@@ -0,0 +1,32 @@
...
+
+    # Set username/password, pgbouncer uses this to authenticate to RDS
+    echo '"'$db_user'"' '"'$db_password'"' > /adaguc/pgbouncer/userlist.txt
+
+    # Rewrite pgbouncer config. $db_host could be an RDS hostname or adaguc-db when running through docker-compose.
+    pgbouncer_cfg="adaguc = host=$db_host port=$db_port dbname=$db_name"
+    sed -i "s/^adaguc = .*/${pgbouncer_cfg}/g" /adaguc/pgbouncer/pgbouncer.ini
+
+    # Rewrite $ADAGUC_DB
+    export ADAGUC_DB="host=localhost port=${db_port} user=${db_user} password=${db_password} dbname=${db_name}"
```
mgrunbauer: This part seems a little bit nasty to me.

`ADAGUC_DB` gets set through `docker compose` or the task definition when running in AWS. I extract the `host=...` section to instruct pgbouncer.
I then rewrite `ADAGUC_DB` to just connect to `host=localhost`.

It works for adaguc-server, it connects to localhost. However if you login to the container `echo $ADAGUC_DB`, it will still show `host=rds...` and not the rewritten variable.

I can't find a way to "rewrite" this environment variable and make docker remember it. Maybe it's not an issue but good to keep in mind.

.github/workflows/docker-publish.yml
```diff
@@ -15,28 +15,6 @@ jobs:
       - name: Check out the repo
         uses: actions/checkout@v4
 
-      - name: Build and potentially push Docker image
```
mgrunbauer: Will revert these.

.github/workflows/docker-publish.yml
```diff
@@ -54,9 +32,9 @@ jobs:
         uses: docker/build-push-action@v5
         with:
           context: .
-          platforms: linux/amd64,linux/arm64
+          platforms: linux/amd64
           # ensure latest base image is used
           pull: true
           push: ${{ github.repository == 'KNMI/adaguc-server' }}
-          tags: ${{ steps.meta.outputs.tags }}
+          tags: ${{ steps.meta.outputs.tags }}-03061000
```
lukas-phaf: ?

.github/workflows/docker-publish.yml
```diff
@@ -54,9 +32,9 @@ jobs:
         uses: docker/build-push-action@v5
         with:
           context: .
-          platforms: linux/amd64,linux/arm64
+          platforms: linux/amd64
```
lukas-phaf: Revert.

Docker/docker-compose.yml
```diff
@@ -20,7 +20,10 @@ services:
         max-size: "200k"
         max-file: "10"
   adaguc-server:
-    image: openearth/adaguc-server
+    build:
+      context: ..
+      dockerfile: ./Dockerfile
```
lukas-phaf: Works without this line, providing the `context` should be enough.

And we probably want to revert this.

Docker/supervisord/adaguc-pgbouncer.conf
```diff
@@ -0,0 +1,16 @@
+[supervisord]
+nodaemon=true
+logfile=/dev/null
```
lukas-phaf, line 3: We silence the logging of supervisord?

Dockerfile
```diff
@@ -93,12 +94,11 @@ COPY tests /adaguc/adaguc-server-master/tests
 COPY runtests.sh /adaguc/adaguc-server-master/runtests.sh
 
 # Run adaguc-server functional and regression tests
-RUN bash runtests.sh
+# RUN bash runtests.sh
```
lukas-phaf: Revert.

Docker/pgbouncer/pgbouncer.ini
```diff
@@ -0,0 +1,12 @@
+[databases]
+adaguc = host=adaguc-db port=5432 dbname=adaguc
+
+[pgbouncer]
+listen_port = 5432
+listen_addr = *
+auth_type = trust
```
lukas-phaf, line 11: On the discussion we had, I think this is the interesting bit:
`PgBouncer handles its own client authentication and has its own database of users. These settings control this.`
From: https://www.pgbouncer.org/config.html

We use `auth_type` is trust, which means:
`No authentication is done. The user name must still exist in auth_file.`

So, anyone claiming to be `adaguc` is accepted by pgbouncer. And pgbouncer handles the login with postgres, using the password as provided in `userlist.txt`.
Let me know if this understanding is not correct.

Docker/pgbouncer/pgbouncer.ini
```diff
@@ -0,0 +1,12 @@
+[databases]
+adaguc = host=adaguc-db port=5432 dbname=adaguc
```
lukas-phaf, line 4: Maybe good to add a comments that this gets overwritten in production.

Maybe for all lines that get changes, also userlist.txt.

Docker/pgbouncer/pgbouncer.ini
```diff
@@ -0,0 +1,12 @@
...
+
+[pgbouncer]
+listen_port = 5432
+listen_addr = *
+auth_type = trust
+auth_file = /adaguc/pgbouncer/userlist.txt
+logfile = pgbouncer.log
+pidfile = pgbouncer.pid
+server_tls_sslmode = require
```
lukas-phaf, line 17: And this gets overwritten locally, correct?

Docker/run_supervisord.sh
lukas-phaf: I don't think it is possible to change environment of the parent shell, so I guess this can not be avoided.

.github/workflows/docker-publish.yml
mgrunbauer: Did this together with Maarten when building images, to keep track of different versions on the same dev branch.

Will revert this and all changes in `docker-publish.yml`, was only meant for testing/debugging.

Docker/pgbouncer/pgbouncer.ini
mgrunbauer: Correct, it gets replaced with `disable`.

Running adaguc local with `server_tls_sslmode = require` did not work, and I did not look into configuring SSL for the local/docker postgres instance.

Docker/pgbouncer/pgbouncer.ini
mgrunbauer: Yes, this is also how I understand it.

Verified this by starting through docker compose, execing into the adaguc container and doing:
```bash
psql "host=localhost port=5432 user=adaguc password=hallo dbname=adaguc"
```
And I get a psql prompt. And then trying the following:
```bash
psql "host=localhost port=5432 user=hallo password=adaguc dbname=adaguc"
```
And I cannot login.

So to summarize:

- pgbouncer connecting to postgres: authentication happens by using `userlist.txt`
- adaguc connecting to pgbouncer: only username must be present in `userlist.txt`

Docker/supervisord/adaguc-pgbouncer.conf
mgrunbauer: Removing this creates a `supervisord.log` in `/adaguc/adaguc-server-master`. It contains:
```
cat supervisord.log
2024-03-06 13:22:04,597 INFO supervisord started with pid 12
2024-03-06 13:22:05,599 INFO spawned: 'adaguc' with pid 13
2024-03-06 13:22:05,600 INFO spawned: 'pgbouncer' with pid 14
2024-03-06 13:22:07,473 INFO success: adaguc entered RUNNING state, process has stayed up for > than 1 seconds (startsecs)
2024-03-06 13:22:07,473 INFO success: pgbouncer entered RUNNING state, process has stayed up for > than 1 seconds (startsecs)
```

However when the option `logfile=/dev/null` is set, the same logging still appears through `docker logs`. Writing to a file also keeps the logging visible through `docker logs`.

I could change it so it writes to a separate log file, although I suspect we would only be able to reach this files when working/debugging locally. What do you suggest?

Docker/pgbouncer/pgbouncer.ini
mgrunbauer: I've added comments :+1:

Docker/supervisord/adaguc-pgbouncer.conf
lukas-phaf: If it's in the docker logs, it is fine.

mgrunbauer:
For reference, the following points were raised during an irl review, and have been fixed:
- Support mixed case "true" in config
- Order independent ADAGUC_DB vars in run_supervisord.sh
- "PGBOUNCER_DISABLE_TLS=true" support / support adaguc-db with TLS
- pgbouncer shouldn't store logfile in container

The following still needs to be verified:
- Verify redirect_stderr=true for adaguc by writing to stderr

maartenplieger (approved):

PR #354, opened 2024-03-12
Added cube call to fix-redis-caching code

ernstdevreede:
This PR adds the EDR cube call to the fix-redis-caching branch


maartenplieger (commented):
Interesting to support EDR cube which such little code.

- How can I review this? Would it work with autowms?
- Please change this MR to point to master.
- Can you add tests? Possibly unit test and a functional test?
- Please add documentation to the documentation section

PR #361, opened 2024-04-08
Optimize getFilesAndIndicesForDimensions

### LINKED ISSUE #341: Simplify query used by `CDBAdapterPostgreSQL::getFilesAndIndicesForDimensions`
The function `CDBAdapterPostgreSQL::getFilesAndIndicesForDimensions()` constructs a query of the form:
```
select distinct * from
    (select a0.path,member,dimmember,height,dimheight,time,dimtime from
        (select path,dimmember,member from t20240301t095153031_xkq9pp0nv3kqmonttvhf where member = 'member4' and adaguctilinglevel != -1 ORDER BY member DESC limit 512)a0
        ,(select path,dimheight,height from t20240301t095153038_hqug9msl1wgxpew4yozf where abs(5000 - height) = (select min(abs(5000 - height)) from t20240301t095153038_hqug9msl1wgxpew4yozf)ORDER BY height DESC)a1
        ,(select path,dimtime,time from t20240301t095153043_oe2pyeutb8pdrvb7gu7r where time = '2017-01-01T00:15:00Z' ORDER BY time DESC)a2
     where a0.path=a1.path and a0.path=a2.path)T
order by member,height,time
```

A simpler from that does (roughly?) the same is:
```
select distinct t1.path,member,dimmember,height,dimheight,time,dimtime from
    t20240301t095153031_xkq9pp0nv3kqmonttvhf t1
inner join t20240301t095153038_hqug9msl1wgxpew4yozf t2
    ON t1.path = t2.path
inner join t20240301t095153043_oe2pyeutb8pdrvb7gu7r t3
    ON t1.path = t3.path
where
    member = 'member4' and
    t1.adaguctilinglevel != -1 and
    abs(5000 - height) = (select min(abs(5000 - height)) from t20240301t095153038_hqug9msl1wgxpew4yozf) and
    time = '2017-01-01T00:15:00Z'
order by member,height,time
limit 512
```

This should give the same result, except for the limit. In the original query it is on the first dimension (which is incorrect, it should be on time). In the second query, it limits the total amount of files returned to 512.

This new query should of course be tested to make sure it give equivalent results.

mgrunbauer:
This MR makes the following changes:
- When requesting data with multiple dimensions, instead of executing queries per dimension, execute a single query for all dimensions. This is happening in `getFilesAndIndicesForDimensions` and `getTableNameForPathFilterAndDimensions`.
- Check if `pathfiltertablelookup` table exists only once during adaguc-server lifetime, instead of once per table access.
- SQLite will now return dim values in the same format than PSQL (e.g. always 10.0 instead of 10)
- You can now run all tests against the postgresql backend.
- The `maxquerylimit` limit now limits the total results from `getFilesAndIndicesForDimensions()` instead of limiting on the first dimension queried. This fixes: https://github.com/KNMI/adaguc-server/issues/341

All changes combined result in less queries. For example `GetMap` required 12 queries before, only 3 queries with this MR.

This slightly improves performance, running the complete benchmark locally takes 140 seconds on master, 128 seconds on this branch.

I've replaced some of the `getTableNameForPathFilterAndDimension()` calls with `getTableNameForPathFilterAndDimensions()` (plural dimension**s**). The former is now a wrapper function that executes the latter, so the code path is identical. Replacing the remaining `getTableNameForPathFilterAndDimension` calls requires some refactoring.

---

Things out of scope for this PR:
- SQLite will not implement the changes made to the PostgreSQL class. SQLite might become deprecated.
- Fix test dependencies when running tests using PostgreSQL. I worked around this with a "temporary band aid" (drop all tables).
- "query hidden" with `SHOW_QUERYINFO` is not required for now.
- Postgres is still using `real` to store floats, which will result in precision loss. We should probably use `double precision`. I've decided to skip it in this MR for now.


adagucserverEC/CDBAdapterPostgreSQL.cpp
```diff
@@ -356,175 +357,110 @@ CDBStore::Store *CDBAdapterPostgreSQL::getFilesAndIndicesForDimensions(CDataSour
...
+        query.printconcat("AND %s = '%s' ", dimName, dimVal);
       }
-      subQuery.printconcat("ORDER BY %s DESC limit %d)a%d ", netCDFDimName.c_str(), limit, i);
-      // subQuery.printconcat("ORDER BY %s DESC )a%d ",netCDFDimName.c_str(),i);
     } else {
-      subQuery.printconcat("ORDER BY %s DESC)a%d ", netCDFDimName.c_str(), i);
+      // Find value within range of dimVals[0] and dimVals[1]
+      // Get closest lowest value to this requested one, or if request value is way below get earliest value:
+      query.printconcat(
+        "AND %s >= (SELECT MAX(%s) FROM %s WHERE %s <= '%s' OR %s = (SELECT MIN(%s) FROM %s)) AND %s <= '%s' ",
```
maartenplieger: To test the functionaltests (bash runtests.sh) with postgres, you can replace globally over all xml files:
```xml
<DataBase dbtype="sqlite" parameters="{ADAGUC_TMP}/adaguc.autoresource.db"/>
```
with 
```xml
<DataBase parameters="{ADAGUC_DB}"/>
```

I noticed differences like:

```xml
<Dimension name="extra" units="extra" default="0.9" multipleValues="1" nearestValue="0" >0.0,0.1,0.2,0.3,0.4,0.5,0.6,0.7,0.8,0.9</Dimension>
```
becomes now:
```xml
<Dimension name="extra" units="extra" default="0.9" multipleValues="1" nearestValue="0" >0,0.1,0.2,0.3,0.4,0.5,0.6,0.7,0.8,0.9</Dimension>
```

mgrunbauer: I think not all errors are the result of this branch. When I run the functional tests on master using postgresql, I get similar errors. Running against sqlite (default) works fine. I've tracked it down to a difference when selecting a `real` from psql vs sqlite.

We currently store all dim values that are a `CDF_FLOAT` or `CDF_DOUBLE` as a `real` in the database.

When a dimension value like `10.0` is stored as real in sqlite, and you `select` it, you get `10.0`.
If you store `10.0` in postgresql, the `select` rounds it to `10`.

We could think about storing these dimension fields as a `numeric (10, 1)`, however I don't know if this fixes all issues or introduces new ones.

This happens when you `select` a real in postgresql:
```
adaguc=# select path, height from t20240420t142720604_h78goam7gcvl33cthua6 where height=10.0;
                                     path                                      | height 
-------------------------------------------------------------------------------+--------
 /adaguc/adaguc-server-master/data/datasets/test_WINS50_windshear_processor.nc |     10
(1 row)

adaguc=# update t20240420t142720604_h78goam7gcvl33cthua6 set height=10.0 where height=10;
UPDATE 1
adaguc=# select path, height from t20240420t142720604_h78goam7gcvl33cthua6 where height=10.0;
                                     path                                      | height 
-------------------------------------------------------------------------------+--------
 /adaguc/adaguc-server-master/data/datasets/test_WINS50_windshear_processor.nc |     10
(1 row)
```
And this is what happens when you select a `real` in sqlite. Second field is the height dimension field, created from the same netcdf. No rounding is done:
```
sqlite> select * from t20240421t140242178_ohzu24f9uf7q1y9k07i1;                                                                                                                                                                                                                           
/adaguc/adaguc-server-master/data/datasets/test_WINS50_windshear_processor.nc|10.0|0|2024-01-18T23:05:50Z|0|-1250.0|-1250.0|541250.0|583750.0|0|1|2|3
/adaguc/adaguc-server-master/data/datasets/test_WINS50_windshear_processor.nc|100.0|1|2024-01-18T23:05:50Z|0|-1250.0|-1250.0|541250.0|583750.0|0|1|2|3
/adaguc/adaguc-server-master/data/datasets/test_WINS50_windshear_processor.nc|200.0|2|2024-01-18T23:05:50Z|0|-1250.0|-1250.0|541250.0|583750.0|0|1|2|3
/adaguc/adaguc-server-master/data/datasets/test_WINS50_windshear_processor.nc|300.0|3|2024-01-18T23:05:50Z|0|-1250.0|-1250.0|541250.0|583750.0|0|1|2|3
/adaguc/adaguc-server-master/data/datasets/test_WINS50_windshear_processor.nc|600.0|4|2024-01-18T23:05:50Z|0|-1250.0|-1250.0|541250.0|583750.0|0|1|2|3

```

mgrunbauer: Resolving this comment:

- The psql vs sqlite issue is now solved by applying an sqlite "formatting hack" to make sqlite and psql consistent.
- I've added a `starttests_psql.sh` which performs the xml replace and runs all tests against psql. It is possible to run it inside Docker as well, if you run adaguc-db through docker compose. I'll describe this part in the documentation.

adagucserverEC/CRequest.cpp
```diff
@@ -1177,6 +1177,20 @@ int CRequest::fillDimValuesForDataSource(CDataSource *dataSource, CServerParams
       }
     }
 
+    // Check if requested time dimensions use valid characters
```
mgrunbauer, line 1180: This check was done in `CDBAdapterPostgreSQL.cpp`, however this is pure input validation. For now I moved it to `CRequest.cpp` since that is where other input validation already happens. Maybe it could help to move all input validation to a separate cpp file.

tests/expectedoutputs/TestDataPostProcessor/test_DataPostProcessor_WindShear_GetFeatureInfo.json
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+[{"name":"output","standard_name":"output","feature_name":"output_0","units":"m/s","point":{"SRS":"EPSG:4326","coords":"3.256556,52.592725"},"dims":["elevation","time"],"data":{"600":{"2019-01-01T22:00:00Z":"-0.195704"}}},{"name":"var_wind_speed_at_600","standard_name":"wind_speed","feature_name":"wind_speed_1","units":"m s-1","point":{"SRS":"EPSG:4326","coords":"3.256556,52.592725"},"dims":["elevation","time"],"data":{"600":{"2019-01-01T22:00:00Z":"16.896723"}}},{"name":"var_wind_speed_at_200","standard_name":"wind_speed","feature_name":"wind_speed_2","units":"m s-1","point":{"SRS":"EPSG:4326","coords":"3.256556,52.592725"},"dims":["elevation","time"],"data":{"600":{"2019-01-01T22:00:00Z":"16.701019"}}}]
```
mgrunbauer, line 1: I've changed the test data so that it is now consistent for both PSQL and Sqlite. The changes are that the dimension value changes from e.g. `10.0` to `10`.

adagucserverEC/CDBAdapterPostgreSQL.cpp
```diff
@@ -356,181 +374,104 @@ CDBStore::Store *CDBAdapterPostgreSQL::getFilesAndIndicesForDimensions(CDataSour
...
   CDBStore::Store *store = NULL;
   try {
     store = DB->queryToStore(query.c_str(), true);
   } catch (int e) {
     if ((CServerParams::checkDataRestriction() & SHOW_QUERYINFO) == false) query.copy("hidden");
-    setExceptionType(InvalidDimensionValue);
+    // The method "setExceptionType" always sets the HTTP statuscode to 404. This breaks the test:
+    //    test_WMSGetCapabilities_no_error_on_existing_dataset_misconfigured_layer
+    // SQLite does not call setExceptionType when the query fails.
+    // setExceptionType(InvalidDimensionValue);
```
mgrunbauer: A test was added in https://github.com/KNMI/adaguc-server/pull/387 which breaks when ran against postgresql. The test asserts that Adaguc should return 200 even if one of the layers is broken.

During the test, a non-existing table gets queried. In SQLite this exception is caught without calling `setExceptionType`. In PSQL we do set `setExceptionType`, which in turn always sets the response HTTP status code to 404.

I think it is better if the database class does not bother with HTTP status codes, maybe this behavior is better handled somewhere else. If we want Adaguc to always respond with a 200, even if a table is missing, we should remove this call.

Dockerfile
```diff
@@ -85,16 +85,21 @@ COPY data /adaguc/adaguc-server-master/data
 COPY python /adaguc/adaguc-server-master/python
 
 ######### Third stage, test ############
-FROM base AS test
+# To run the tests against a postgres db, see docs/test_postgesql.md
+FROM base as test
```
maartenplieger: Please keep as as uppercase (as -> AS) to limit docker warnings

Dockerfile
```diff
@@ -85,16 +85,21 @@ COPY data /adaguc/adaguc-server-master/data
...
 RUN pip install --no-cache-dir -r requirements-dev.txt
 
 COPY tests /adaguc/adaguc-server-master/tests
 COPY runtests.sh /adaguc/adaguc-server-master/runtests.sh
+COPY runtests_psql.sh /adaguc/adaguc-server-master/runtests_psql.sh
 
-# Run adaguc-server functional and regression tests
+# Run adaguc-server functional and regression tests. See also `./doc/developing/testing.md`
 RUN bash runtests.sh
+# RUN bash runtests_psql.sh
```
maartenplieger, line 102: Curious to know if this could work in github actions? That would be nice! In the future we could get rid of sqlite completely.

adagucserverEC/CDBAdapterPostgreSQL.cpp
```diff
@@ -356,181 +374,104 @@ CDBStore::Store *CDBAdapterPostgreSQL::getFilesAndIndicesForDimensions(CDataSour
...
+      if (requestedDimVals[i].equals("*")) continue;
+
+      if (i != 0) {
+        whereStatement.printconcat(" OR ");
       }
-      if (dataSource->queryBBOX) {
 
-        subQuery.printconcat("adaguctilinglevel = %d and minx >= %f and maxx <= %f and miny >= %f and maxy <= %f ", dataSource->queryLevel, dataSource->nativeViewPortBBOX[0],
-                             dataSource->nativeViewPortBBOX[2], dataSource->nativeViewPortBBOX[1], dataSource->nativeViewPortBBOX[3]);
+      CT::string *dimVals = requestedDimVals[i].splitToArray("/");
```
maartenplieger: BTW, I think we should think of CT::string split options. splitToStack will create a std::vector<CT::string>.

adagucserverEC/CDBAdapterPostgreSQL.cpp
```diff
@@ -356,181 +374,104 @@ CDBStore::Store *CDBAdapterPostgreSQL::getFilesAndIndicesForDimensions(CDataSour
...
+        // Find value within range of dimVals[0] and dimVals[1]
+        // Get closest lowest value to this requested one, or if request value is way below get earliest value:
+        whereStatement.printconcat("%s >= (SELECT MAX(%s) FROM %s WHERE %s <= '%s' OR %s = (SELECT MIN(%s) FROM %s)) AND %s <= '%s'", dimName, dimName, tableName, dimName, dimVals[0].c_str(), dimName,
+                                   dimName, tableName, dimName, dimVals[1].c_str());
       }
-      subQuery.printconcat("ORDER BY %s DESC limit %d)a%d ", netCDFDimName.c_str(), limit, i);
-      // subQuery.printconcat("ORDER BY %s DESC )a%d ",netCDFDimName.c_str(),i);
-    } else {
-      subQuery.printconcat("ORDER BY %s DESC)a%d ", netCDFDimName.c_str(), i);
+      delete[] dimVals;
```
maartenplieger: If we would use splitToStack the delete is less critical if forgotten

adagucserverEC/CDBAdapterPostgreSQL.cpp
maartenplieger: Agree.

adagucserverEC/CDBAdapterPostgreSQL.cpp
```diff
@@ -651,148 +586,187 @@ int CDBAdapterPostgreSQL::autoUpdateAndScanDimensionTables(CDataSource *dataSour
...
-      throw(1);
-    }
-    if (rec->getSize() > 0) {
-      tableName.copy(rec->getRecord(0)->get(3));
-      if (tableName.length() > 0) {
-        lookupTableIsAvailable = true;
-      }
-    }
+    // Generate a new random table name
+    CT::string tableName;
```
maartenplieger: Would be nice to extract generateRandomTableName() to a function.

adagucserverEC/CDBAdapterSQLLite.cpp
```diff
@@ -183,7 +183,16 @@ CDBStore::Store *CDBAdapterSQLLite::CSQLLiteDB::queryToStore(const char *pszQuer
 
   for (size_t rowNumber = 0; rowNumber < numRows; rowNumber++) {
     CDBStore::Record *record = new CDBStore::Record(colModel);
-    for (size_t colNumber = 0; colNumber < numCols; colNumber++) record->push(colNumber, queryValues[colNumber + rowNumber * numCols].c_str());
+    for (size_t colNumber = 0; colNumber < numCols; colNumber++) {
+      // SQlite always returns `10.0` if you insert `10.0` or `10`. PSQL always returns `10`
```
maartenplieger, line 187: Nice. Lets discuss if we can get away with removing sqlite in a next update.

adagucserverEC/CRequest.cpp
maartenplieger: Can you write an issue in github for this?

adagucserverEC/CRequest.cpp
```diff
@@ -1177,6 +1177,20 @@ int CRequest::fillDimValuesForDataSource(CDataSource *dataSource, CServerParams
...
+    for (auto &dim : dataSource->requiredDims) {
+      // FIXME: checkTimeFormat used to get called on every dim value, not just datetime. Check if this is required
+      if (!dim->isATimeDimension) continue;
+
+      CT::string *dimValues = dim->value.splitToArray(",");
+      for (size_t i = 0; i < dimValues->count; i++) {
+        if (!CServerParams::checkTimeFormat(dimValues[i])) {
+          CDBError("Queried dimension %s=%s failed datetime regex", dim->name.c_str(), dim->value.c_str());
+          throw InvalidDimensionValue;
```
maartenplieger, line 1189: To discuss: I know we do it at many places, but we might consider other methods to notify errors instead of throwing exceptions. I understood that raising exceptions is quite performance heavy.

doc/developing/testing.md
```diff
@@ -0,0 +1,70 @@
...
+The build will fail if any of the tests have failed.
+
+
+## Running tests outside Docker via SQLite
+
+Running the tests outside of Docker is possible by executing the `./runtests.sh` script.
+
+> **⚠️ Note**: some tests might fail due to platform architecture differences between amd64 and arm64.
+
+## Using postgres
```
maartenplieger, line 28: This is nice documentation, Good job! Do you think we can make this work with docker in github actions?

tests/starttests_psql.sh
```diff
@@ -0,0 +1,58 @@
...
+# Unfortunately macos does not use GNU sed, so the sed command differs slightly (-i '' vs -i)
+project_root_dir=`dirname -- "$( readlink -f -- "$0"; )"`
+project_root_dir+="/../data/config"
+if [ -z "${project_root_dir}" ]; then
+    echo "Could not determine root dir of project!"
+    exit 1
+fi
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    find "${project_root_dir}" -type f -name '*.xml' | xargs sed -i '' 's;<DataBase dbtype="sqlite" parameters="{ADAGUC_TMP}/adaguc.autoresource.db"/>;<DataBase parameters="{ADAGUC_DB}"/>;g'
```
maartenplieger: Would it not be better to just fix the config files where this is set with the ADAGUC_DB parameter?

maartenplieger (commented):
Nice!!!!! I am deploying a dev version for geoweb to spot any issues.

adagucserverEC/CRequest.cpp
mgrunbauer: Created https://github.com/KNMI/adaguc-server/issues/398

adagucserverEC/CDBAdapterPostgreSQL.cpp
mgrunbauer: I'll remove `    // setExceptionType(InvalidDimensionValue);`

doc/developing/testing.md
mgrunbauer: I'll make an issue to see if this is possible.

tests/starttests_psql.sh
mgrunbauer: Yes, this works great! :+1: I'll change this.

adagucserverEC/CRequest.cpp
mgrunbauer: Agree!

doc/developing/testing.md
mgrunbauer: Created issue: https://github.com/KNMI/adaguc-server/issues/400

Dockerfile
mgrunbauer: See https://github.com/KNMI/adaguc-server/issues/400

maartenplieger (approved):

PR #362, opened 2024-04-10
Support INT64 in CDFDataWriter

lukas-phaf:


adagucserverEC/CNetCDFDataWriter.cpp
```diff
@@ -381,13 +381,20 @@ int CNetCDFDataWriter::init(CServerParams *srvParam, CDataSource *dataSource, in
...
+            ((unsigned long *)var->data)[j] = dimValue.toLong();
+            break;
           case CDF_FLOAT:
             ((float *)var->data)[j] = dimValue.toFloat();
             break;
           case CDF_DOUBLE:
             ((double *)var->data)[j] = dimValue.toDouble();
             break;
           default:
+            CDBError("Unknown var type [%d]", var->getType());
```
mgrunbauer: Maybe also log the dimension name (`dimName.c_str()`) so that you know what dimension the error relates to?

mgrunbauer (dismissed):

adagucserverEC/CNetCDFDataWriter.cpp
lukas-phaf: Good idea! Done.

maartenplieger (approved):

PR #367, opened 2024-04-18
Add cube to redis

ernstdevreede:


ernstdevreede:
Remarks from @maartenplieger  

Interesting to support EDR cube which such little code.

- How can I review this? Would it work with autowms?
- Please change this MR to point to master.
- Can you add tests? Possibly unit test and a functional test?
- Please add documentation to the documentation section

PR #373, opened 2024-05-22
EDR Metadata can now be harmonized

### LINKED ISSUE #359: Improve EDR metadata for parameters
Currently the EDR service returns the following for parameters (i18n was removed??):
(From https://geoservices.knmi.nl/edr/collections )

```json
{
  "air_temperature_hagl": {
    "type": "Parameter",
    "id": "air_temperature_hagl",
    "label": "air_temperature_hagl",
    "unit": {
      "symbol": "C"
    },
    "observedProperty": {
      "id": "air_temperature_hagl",
      "label": "air_temperature_hagl"
    }
  }
}
```

Could it be adjusted towards:
```json
{
  "air_temperature_hagl": {
    "type": "Parameter",
    "description":"HAP1(NL) Air temperature at 2 m",
    "id": "air_temperature_hagl",
    "label": "air_temperature_hagl",
    "unit": {
      "label": "degree Celsius",
      "symbol": {
        "value": "°C",
        "type": "http://www.opengis.net/def/uom/UCUM/"
      }
    },
    "observedProperty": {
      "id": "https://vocab.nerc.ac.uk/standard_name/air_temperature/",
      "label": "Air temperature"
    }
  }
}
```

Question: Do we really need  `id`, `label` and `unit.label` ?
We need to check the covjson spec on the distinction between `id`, `label` and `description`

Description could be obtained from the WMS title in the WMS GetCapabilities
Suggested EDR config could be:
```xml
 <EdrParameter name="air_temperature_hagl" unit="°C" standard_name="air_temperature" label="Air temperature, 2 metre"/>
 ```
(Label TBD)

### PR DESCRIPTION
**Pull request by maartenplieger**
Closes https://github.com/KNMI/adaguc-server/issues/359

This allows to configure extra metadata for the parameters in the Edr collection response

```xml
    <OgcApiEdr>
        <EdrCollection name="harmonie">
            <EdrParameter
                name="air_temperature__at_2m"
                unit="°C"
                standard_name="air_temperature"
                observed_property_label="Air temperature"
                parameter_label="Air temperature, 2 metre" />
        </EdrCollection>
    </OgcApiEdr>
```

# To test:

- Follow instructions as in doc/tutorials/Configure_EDR_service.md
- Start the server and visit http://localhost:8080/edr/collections

The parameter in the collection endpoint now lists:

```json
"parameter_names": {
    "air_temperature__at_2m": {
      "type": "Parameter",
      "id": "air_temperature__at_2m",
      "label": "Air temperature, 2 metre",
      "description": "harmonie - air_temperature__at_2m (air_temperature__at_2m)",
      "unit": {
        "symbol": {
          "value": "°C",
          "type": "http://www.opengis.net/def/uom/UCUM"
        }
      },
      "observedProperty": {
        "id": "https://vocab.nerc.ac.uk/standard_name/air_temperature",
        "label": "Air temperature"
      }
    }
  }
```


python/python_fastapi_server/routers/edr.py
```diff
@@ -420,11 +471,16 @@ def get_params_for_collection(edr_collection: str) -> dict[str, Parameter]:
...
+        # If standard name was configured, use that instead with a vocabulary
+        if "standard_name" in param_el:
+            param_id = VOCAB_ENDPOINT_URL + param_el["standard_name"]
         param = Parameter(
             id=param_el["name"],
-            observedProperty=ObservedProperty(id=param_el["name"], label=label),
+            observedProperty=ObservedProperty(id=param_id, label=label),
             type="Parameter",
-            unit=Unit(symbol=param_el["unit"]),
+            unit=Unit(symbol=symbol),
```
lukas-phaf: So, for now we don't put a label in the `unit` (as there is in the example in #359).

lukas-phaf: These are breaking changes. Are the users (e.g. Geoweb) okay with that?

python/python_fastapi_server/routers/edr.py
```diff
@@ -420,11 +471,16 @@ def get_params_for_collection(edr_collection: str) -> dict[str, Parameter]:
...
 
+        symbol = Symbol(value=param_el["unit"], type=SYMBOL_TYPE_URL)
+        param_id = param_el["name"]
+        # If standard name was configured, use that instead with a vocabulary
+        if "standard_name" in param_el:
+            param_id = VOCAB_ENDPOINT_URL + param_el["standard_name"]
         param = Parameter(
             id=param_el["name"],
-            observedProperty=ObservedProperty(id=param_el["name"], label=label),
+            observedProperty=ObservedProperty(id=param_id, label=label),
```
lukas-phaf: We are using the same `label` in the `ObservedProperty` and the `Parameter` technically this is not correct. The property is an "air temperature", the parameter is "air temperature measured at 2 m altitude with sensor X".

I think the label provided in the example config is appropriate for the `parameter`, but not for the `ObservedProperty`.

python/python_fastapi_server/routers/edr.py
```diff
@@ -420,11 +471,16 @@ def get_params_for_collection(edr_collection: str) -> dict[str, Parameter]:
         else:
             label = param_el["name"]
 
+        symbol = Symbol(value=param_el["unit"], type=SYMBOL_TYPE_URL)
+        param_id = param_el["name"]
+        # If standard name was configured, use that instead with a vocabulary
+        if "standard_name" in param_el:
+            param_id = VOCAB_ENDPOINT_URL + param_el["standard_name"]
         param = Parameter(
             id=param_el["name"],
```
lukas-phaf: So we are not yet adding a `description`, like in the example in #359.

adagucserverEC/Definitions.h
```diff
@@ -28,7 +28,7 @@
 #ifndef Definitions_H
 #define Definitions_H
 
-#define ADAGUCSERVER_VERSION "2.21.2" // Please also update in the Dockerfile to the same version
+#define ADAGUCSERVER_VERSION "2.21.3" // Please also update in the Dockerfile to the same version
```
lukas-phaf: Big enough version bump for "breaking change" of EDR output?

python/python_fastapi_server/routers/edr.py
```diff
@@ -123,12 +157,29 @@ def init_edr_collections(adaguc_dataset_dir: str = os.environ["ADAGUC_DATASET_DI
...
                         ):
-                            edr_params.append(
-                                {
-                                    "name": edr_parameter.attrib.get("name"),
+                            name = edr_parameter.attrib.get("name")
+                            layer = find_layer_configuration(
+                                adaguc_dataset_layers, name
+                            )
+                            # If the layer is present in the adaguc dataset configuration, continue.
+                            if layer is not None:
```
lukas-phaf: Do we need to throw and error of no layer is matched, as the EDR line is now silently ignored?

doc/tutorials/Configure_EDR_service.md
```diff
@@ -23,14 +23,14 @@ This file is available in the adaguc-server repository with location `data/datas
...
-<?xml version="1.0" encoding="UTF-8" ?>
+
 <Configuration>
 
     <OgcApiFeatures/>
 
     <OgcApiEdr>
         <EdrCollection name="harmonie">
-            <EdrParameter name="air_temperature__at_2m" unit="Celsius"/>
+            <EdrParameter name="air_temperature__at_2m" unit="°C"  standard_name="air_temperature" label="Air temperature, 2 metre"/>
```
lukas-phaf: If we are matching up a layer anyway, why not use the `units` from the layer variable. I guess the (?) reason is that the layer variable `units` is optional in this config file, and we don't want to do a call into adaguc C++ when getting back `/collections`?

python/python_fastapi_server/routers/edr.py
```diff
@@ -420,11 +471,16 @@ def get_params_for_collection(edr_collection: str) -> dict[str, Parameter]:
         else:
             label = param_el["name"]
 
+        symbol = Symbol(value=param_el["unit"], type=SYMBOL_TYPE_URL)
+        param_id = param_el["name"]
+        # If standard name was configured, use that instead with a vocabulary
+        if "standard_name" in param_el:
+            param_id = VOCAB_ENDPOINT_URL + param_el["standard_name"]
         param = Parameter(
```
lukas-phaf: This is the `Parameter` object that is used in the metadata in the `/collections` endpoint. But if you do a data query, you also get a `Parameter` object back which has a slightly different type (long story...). This is constructed on line 1056 of edr.py. I think we want to update that as well?

python/python_fastapi_server/routers/edr.py
```diff
@@ -420,11 +471,16 @@ def get_params_for_collection(edr_collection: str) -> dict[str, Parameter]:
         else:
             label = param_el["name"]
 
+        symbol = Symbol(value=param_el["unit"], type=SYMBOL_TYPE_URL)
```
lukas-phaf: From the diff, I guess there is not currently a test which shows/tests the result of this?

python/python_fastapi_server/routers/edr.py
maartenplieger: Ah, good point. I would like to follow #359. I think I forgot the label in the unit.

maartenplieger: Yes, it will be allright since the KDP edr api does the same.

python/python_fastapi_server/routers/edr.py
maartenplieger: Can we use the standard_name (`air_temperature`) in the label of `observedProperty` instead, or should we support an additional configuration option?

Probably an additional config option?

python/python_fastapi_server/routers/edr.py
maartenplieger: No indeed. This could be the title of the WMS Layer I presume?

python/python_fastapi_server/routers/edr.py
maartenplieger: It is now silently ignored. What would be the better option...... I guess throwing an error is very clear?

doc/tutorials/Configure_EDR_service.md
maartenplieger: It is indeed optional. So we cannot use that. Unless we start reading the GetCapabilities document.......

python/python_fastapi_server/routers/edr.py
maartenplieger: Ah yes!

python/python_fastapi_server/routers/edr.py
maartenplieger: Indeed. It would be good to add a test, lets see if that is feasible.

python/python_fastapi_server/routers/edr.py
maartenplieger: Added the label.

adagucserverEC/Definitions.h
maartenplieger: Changed to 2.22.0

doc/tutorials/Configure_EDR_service.md
maartenplieger: This information is not avaiable in the GetCapabilities document. Created an issue https://github.com/KNMI/adaguc-server/issues/374

python/python_fastapi_server/routers/edr.py
maartenplieger: A warning is printed instead.

python/python_fastapi_server/routers/edr.py
maartenplieger: Lets write those in https://github.com/KNMI/adaguc-server/issues/375

NEWS.md
```diff
@@ -1,3 +1,8 @@
+**Version 2.22.0 2024-05-22**
+- EDR Metadata can now be harmonized: https://github.com/KNMI/adaguc-server/issues/359. 
+- See  [Configure_EDR_service](doc/tutorials/Configure_EDR_service.md) for details.
```
lukas-phaf:
```suggestion
- See [Configure_EDR_service](doc/tutorials/Configure_EDR_service.md) for details.
```

NEWS.md
```diff
@@ -1,3 +1,8 @@
+**Version 2.22.0 2024-05-22**
+- EDR Metadata can now be harmonized: https://github.com/KNMI/adaguc-server/issues/359.
```
lukas-phaf: `Harmonized` doesn't mean much to me. What do you mean?

doc/tutorials/Configure_EDR_service.md
```diff
@@ -43,18 +48,63 @@ Create the following file at the filepath `$ADAGUC_DATASET_DIR/edr.xml`. You can
...
+                standard_name="air_temperature"
+                observed_property_label="Air temperature"
+                parameter_label="Air temperature, 2 metre" />
+```                
+
+- name: Mandatory, Should be one of the WMS Layer names as advertised in the WMS GetCapabilities
+- unit: Mandatory, Sets the unit for the parameter in the parameter_names section of the collection document
+- standard_name: Sets the observedProperty id
+- observed_property_label: Sets the observedProperty label
+- parameter_label: Sets the label for the parameter in the parameter_names section
```
lukas-phaf: If these are not mandatory, where does the data come from?

python/python_fastapi_server/routers/edr.py
```diff
@@ -126,12 +132,41 @@ def init_edr_collections(adaguc_dataset_dir: str = os.environ["ADAGUC_DATASET_DI
...
+                            if "standard_name" in edr_parameter.attrib:
+                                edr_param["standard_name"] = (
+                                    edr_parameter.attrib.get("standard_name")
+                                )
+                                # If there is a standard name, set the label to the same as fallback
+                                edr_param["observed_property_label"] = (
+                                    edr_parameter.attrib.get("standard_name")
+                                )
+
+                            # Try to take the oobserved_property_label from the configuration
```
lukas-phaf:
```suggestion
                            # Try to take the observed_property_label from the configuration
```

python/python_fastapi_server/routers/edr.py
```diff
@@ -440,28 +475,82 @@ async def get_collectioninfo_for_id(
...
+        _type_: parameter element_
+    """
+    edr_collections = get_edr_collections()
+    edr_collection_parameters= edr_collections[edr_collection_name]["parameters"]
+    for param_el in edr_collection_parameters:
+        if param_id == param_el["name"]:
+            return param_el
+    return None
+
+def get_param_metadata( param_id:str, edr_collection_name:str,wmslayers:dict=None)->dict:
```
lukas-phaf:
```suggestion
def get_param_metadata(param_id:str, edr_collection_name:str, wmslayers:dict=None)->dict:
```

python/python_fastapi_server/routers/edr.py
```diff
@@ -440,28 +475,82 @@ async def get_collectioninfo_for_id(
...
-        parameter_names[param_el["name"]] = param
+            
+            param_metadata = get_param_metadata(param_id, edr_collection, wmslayers)
+            logger.info(param_id)
+            logger.info(param_metadata["wms_layer_name"])
+            param = Parameter(
+                id=param_metadata["wms_layer_name"],
+                observedProperty=ObservedProperty(id=param_metadata["observed_property_id"], label=param_metadata["observed_property_label"]),
+                description=param_metadata["wms_layer_title"],
+                type="Parameter",
```
lukas-phaf, line 538: You don't need to provide the `type` field, it is automatically filled in.

python/python_fastapi_server/routers/edr.py
```diff
@@ -1164,7 +1264,7 @@ def covjson_from_resp(dats, vertical_name):
             ]
             domain = Domain(domainType=domain_type, axes=axes, referencing=referencing)
             covjson = Coverage(
-                id="test",
+                id=f"coverage_{(len(covjson_list)+1)}",
```
lukas-phaf, line 1253: `id` is optional. Probably better to not provide it than some random string?

python/python_fastapi_server/routers/edr.py
```diff
@@ -440,28 +475,82 @@ async def get_collectioninfo_for_id(
...
+            "observed_property_id":observed_property_id, 
+            "observed_property_label":observed_property_label,
+            "parameter_label":parameter_label, 
+            "parameter_unit":parameter_unit}
+
+def get_params_for_collection(edr_collection: str, wmslayers: dict) -> dict[str, Parameter]:
     """
     Returns a dictionary with parameters for given EDR collection
     """
     parameter_names = {}
```
lukas-phaf, line 524: Strange name for a dict that holds name -> Parameter pairs.

python/python_fastapi_server/routers/edr.py
```diff
@@ -412,7 +447,7 @@ async def get_collectioninfo_for_id(
     else:
         data_queries = DataQueries(position=EDRQuery(link=position_link))
 
-    parameter_names = get_params_for_collection(edr_collection=edr_collection)
+    parameter_names = get_params_for_collection(edr_collection=edr_collection, wmslayers=wmslayers)
```
lukas-phaf, line 450: Unclear name for this variable. It's a dict that holds more than names.

python/python_fastapi_server/routers/edr.py
```diff
@@ -440,28 +475,82 @@ async def get_collectioninfo_for_id(
 
     return collection, ttl
 
+def get_param_el(edr_collection_name:str, param_id:str):
```
lukas-phaf: What does the `el` stand for?

python/python_fastapi_server/routers/edr.py
```diff
@@ -440,28 +475,82 @@ async def get_collectioninfo_for_id(
...
+
+    Args:
+        edr_collection_name (str): edr collection name
+        param_id (str): parameter id
+
+    Returns:
+        _type_: parameter element_
+    """
+    edr_collections = get_edr_collections()
+    edr_collection_parameters= edr_collections[edr_collection_name]["parameters"]
```
lukas-phaf:
```suggestion
    edr_collection_parameters = edr_collections[edr_collection_name]["parameters"]
```

python/python_fastapi_server/routers/edr.py
```diff
@@ -440,28 +475,82 @@ async def get_collectioninfo_for_id(
...
     parameter_names = {}
     edr_collections = get_edr_collections()
     for param_el in edr_collections[edr_collection]["parameters"]:
-        # Use name as default for label
-        if "label" in param_el:
-            label = param_el["label"]
+
+        param_id = param_el["name"]
+
+
```
lukas-phaf: Lots of empty lines?

python/python_fastapi_server/routers/edr.py
lukas-phaf: This functions is sometimes called with `wmslayers` (in metadata endpoints, I think), and sometimes without (in data endpoints?). This means the Parameters in both can/will be different. Is that what we want?

NEWS.md
maartenplieger: Solved

doc/tutorials/Configure_EDR_service.md
maartenplieger: Described this in more detail

python/python_fastapi_server/routers/edr.py
maartenplieger: It is just the list of wmslayers from the getcapabilities with names and dimensions. Not sure what a better name would be.

python/python_fastapi_server/routers/edr.py
maartenplieger: Changed to get_parameter_config

python/python_fastapi_server/routers/edr.py
maartenplieger: For now I removed the description. Otherwise we need to fetch the getcapabilities for everything. Not sure if we really want the description at this stage.

ernstdevreede (approved):

PR #379, opened 2024-06-10
DRAFT: Fix for reading autowms links with the source= parameter

maartenplieger:


adagucserverEC/CRequest.cpp
```diff
@@ -2460,6 +2459,10 @@ int CRequest::process_querystring() {
     if (equalPos != -1) {
       uriKeyUpperCase = parameters[j].substring(0, equalPos);
       uriValue = parameters[j].c_str() + equalPos + 1;
+      // Remove trailing equals sign, for gracefully supporting invalid decoded urls from adaguc-viewer
+      if (uriValue.endsWith("=")) {
```
maartenplieger, line 2463: As discussed, this is not a fix we want in the adaguc code.

PR #385, opened 2024-07-02
DRAFT: Added x-forwarded-for header in accesslog

maartenplieger:


python/python_fastapi_server/main.py
```diff
@@ -29,7 +29,7 @@
 
 # Set uvicorn access log format using middleware
 access_log_format = (
-    'accesslog %(h)s ; %(t)s ; %(H)s ; %(m)s ; %(U)s ; %(q)s ; %(s)s ; %(M)s ; "%(a)s"'
+    'accesslog %(h)s ; %(t)s ; %(H)s ; %(m)s ; %(U)s ; %(q)s ; %(s)s ; %(M)s ; "%(a)s" ; %({x-forwarded-for}i)s'
```
maartenplieger, line 32: Add " tokens in case the string is empty

PR #395, opened 2024-09-02
Fix: WCS DescribeCoverage does not show dimensions for a multidim layer

belentorrente:
This MR covers a fix to show the values of multiple dimensions in DescribeCoverage, but also fixes an adjacent problem that was found (inconsistent time interval calculations). 

The code to calculate intervals has been made modular, and it can work with leap years and data missing some timesteps. 

Note: Tests fail in one of the Docker tasks due to EPSG:28992 envelopes. This does not happen in the other Docker task. The differences are minimal (likely caused by different versions of proj4?). So the contents of EPSG:28992 envelopes have been ignored when comparing the XML files.


adagucserverEC/CXMLGen.cpp
```diff
@@ -41,6 +43,35 @@
 const char *CXMLGen::className = "CXMLGen";
 int CXMLGen::WCSDescribeCoverage(CServerParams *srvParam, CT::string *XMLDocument) { return OGCGetCapabilities(srvParam, XMLDocument); }
 
+// Function to parse a string to double if numeric
+double parseNumeric(std::string const &str, bool &isNumeric) {
```
maartenplieger: Can these functions be moved to helper files?

adagucserverEC/CXMLGen.cpp
```diff
@@ -775,6 +806,79 @@ int CXMLGen::getWCS_1_0_0_Capabilities(CT::string *XMLDoc, std::vector<MetadataL
   return 0;
 }
 
+void generateRangeSet(CT::string *XMLDoc, MetadataLayer *layer) {
```
maartenplieger: This is specific to WCS right? Maybe mention this in the name?

adagucserverEC/CXMLGen.cpp
belentorrente: Yes, maybe to CXMLGenUtils? Since it is used in CXMLGen.

maartenplieger (dismissed):

maartenplieger (approved):

PR #396, opened 2024-09-02
Testing data explorer

maartenplieger:
To test:

```
git checkout adaguc-data-exploer
cd Docker
docker-compose up -d --build
```

You can visit /explorer

### DIFF
--- a/Docker/docker-compose.yml
+++ b/Docker/docker-compose.yml
@@ -1,5 +1,3 @@
-version: "3.1"
-
 services:
   adaguc-viewer:
     image: openearth/adaguc-viewer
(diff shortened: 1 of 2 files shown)


belentorrente:
Most of the functionality that's implemented works very well. 

Two things I have observed:

- When the legend is not available/not applicable (as in the case of a liveupdate dataset), an "image not found" icon is displayed, as in here:  
image

- Refreshing does not work well, it goes from this:
image

to this: 
image

mgrunbauer:
Nice work!

- I couldn't get the animation to play, it stops at the second step. I can manually step through the timesteps. I think it would be nice to be able to play through all the available time steps.
- Cannot change projection (or I just could not find it)
- Cannot change base layer (or I just could not find it)

ernstdevreede (approved):

PR #397, opened 2024-09-02
Enable trivy vulnerability scanning

lukas-phaf:
Trivy is a deep scanner that can detect both OS en Debian vulnerabilities when scanning the docker image:
https://github.com/aquasecurity/trivy

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -31,7 +31,7 @@ jobs:
           exit-code: '1'
           ignore-unfixed: true
           vuln-type: 'os,library'
-          trivyignores: .trivyignore'
+          trivyignores: .trivyignore
 #          severity: 'CRITICAL,HIGH'
   docker-build-and-publish:
     name: Build Docker image and potentially push to Docker Hub
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -22,6 +22,16 @@ jobs:
           # ensure latest base image is used
           pull: true
           push: false
+          tags: adaguc-server
+      - name: Run Trivy vulnerability scanner
+        uses: aquasecurity/trivy-action@0.24.0
+        with:
+          image-ref: 'adaguc-server'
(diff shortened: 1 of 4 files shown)


.github/workflows/docker-publish.yml
```diff
@@ -22,6 +22,17 @@ jobs:
...
+      - name: Run Trivy vulnerability scanner
+        uses: aquasecurity/trivy-action@0.24.0
+        with:
+          image-ref: 'adaguc-server'
+          format: 'table'
+          exit-code: '1'
+          ignore-unfixed: true
+          vuln-type: 'os,library'
+          trivyignores: .trivyignore
+#          severity: 'CRITICAL,HIGH'
```
mgrunbauer, line 35: What does `severity` do, does it set a threshold and fails if exceeded? What is the default?
If `trivy` finds a vulnerability, does this fail the pipeline?

mgrunbauer (commented):
I see the logging also contains lines like:

> [python] License acquired from METADATA classifiers may be subject to additional terms

Looks like trivy parses license info from the python dependencies. Do we need to look into these messages in this PR?

.github/workflows/docker-publish.yml
lukas-phaf: I haven't look at the license stuff in detail. I think these messages are related to [license scanning](https://aquasecurity.github.io/trivy/v0.33/docs/licenses/scanning/), but not 100% sure. We don't have that turned on at the moment.

lukas-phaf: `severity`: severities of security issues to be displayed (UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL) (default [UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL]

If trivy finds a vulnerability, the build fails (this is the `exit-code`) option.
An example is [here](https://github.com/KNMI/adaguc-server/actions/runs/10664986837/job/29557426021).

mgrunbauer (approved):

PR #399, opened 2024-09-02
Edrfixes - units

ernstdevreede:
Adds some metadata for units in CovJSON.


python/python_fastapi_server/routers/edr_covjson.py
```diff
@@ -76,7 +76,8 @@ def covjson_from_resp(dats, vertical_name, custom_name, collection_name):
             symbol = CovJsonSymbol(
                 value=param_metadata["parameter_unit"], type=SYMBOL_TYPE_URL
             )
-            unit = CovJsonUnit(symbol=symbol)
+            unit_label = {"en": param_metadata["parameter_unit"]}
+            unit = CovJsonUnit(symbol=symbol, label=unit_label)
```
mgrunbauer, line 80: The test `test_collections` in `test_ogc_api_edr.py` calls `/edr/collections`, but the response does not contain the new `"en": ...` label. Is this test still correct?

mgrunbauer (commented):
Could you add a test that asserts that the parameter description for a `position` and `cube` call are identical?

maartenplieger (approved):

Issue #350, opened 2024-03-06
GetCapabilities returns CRS: EPSG:4326 twice


mgrunbauer:
Performing a `GetCapabilities` request returns the `<BoundingBox ...>` element twice for CRS `EPSG:4326` for every layer:

```bash
curl -s "https://geoservices.knmi.nl/adagucserver?dataset=knmi_hap1_nl_2p5km&=&service=WMS&request=GetCapabilities" | grep -e '<BoundingBox CRS="EPSG:4326" minx="48.988500" miny="-0.018500" maxx="55.888500" maxy="11.081500" />' -e 'Title>HAP1'
```
outputs:
```bash
...
<Title>HAP1(NL) Absolute updraft helicity in atmosphere</Title>
<BoundingBox CRS="EPSG:4326" minx="48.988500" miny="-0.018500" maxx="55.888500" maxy="11.081500" />
<BoundingBox CRS="EPSG:4326" minx="48.988500" miny="-0.018500" maxx="55.888500" maxy="11.081500" />
<Title>HAP1(NL) Air pressure at mean sea level</Title>
<BoundingBox CRS="EPSG:4326" minx="48.988500" miny="-0.018500" maxx="55.888500" maxy="11.081500" />
<BoundingBox CRS="EPSG:4326" minx="48.988500" miny="-0.018500" maxx="55.888500" maxy="11.081500" />
...
```

The `<CRS>` and `<BoundingBox ...>` elements also appear twice in the "root" layer:
```
<Layer>
<Title>WMS of  knmi_hap1_nl_2p5km</Title>
<CRS>EPSG:3411</CRS>
<CRS>EPSG:3412</CRS>
<CRS>EPSG:3575</CRS>
<CRS>EPSG:3857</CRS>
<CRS>EPSG:4258</CRS>
<CRS>EPSG:4326</CRS>
<CRS>CRS:84</CRS>
<CRS>EPSG:25831</CRS>
<CRS>EPSG:25832</CRS>
<CRS>EPSG:28992</CRS>
<CRS>EPSG:7399</CRS>
<CRS>EPSG:50001</CRS>
<CRS>EPSG:54030</CRS>
<CRS>EPSG:32661</CRS>
<CRS>EPSG:40000</CRS>
<CRS>EPSG:900913</CRS>
<CRS>EPSG:3067</CRS>
<CRS>EPSG:4326</CRS>
[...]

PR #401, opened 2024-09-03
Store layer metadata

maartenplieger:


adagucserverEC/CDBAdapterPostgreSQL.cpp
```diff
@@ -1080,3 +1080,84 @@ int CDBAdapterPostgreSQL::addFilesToDataBase() {
...
+    if (layerMetaDataStore->size() == 0) {
+      CDBDebug("No results \"%s\"", query.c_str());
+      throw(__LINE__);
+    }
+  }
+
+  auto records = layerMetaDataStore->getRecords();
+  for (auto record : records) {
+    if (record->get("layername")->equals(layerName) && record->get("metadatakey")->equals(metadataKey)) {
+      return record->get("blob");
```
mgrunbauer: Could you add the following before the `return`?
```c++
#ifdef MEASURETIME
  StopWatch_Stop("<CDBAdapterPostgreSQL::getLayerMetadata");
#endif
```

adagucserverEC/CDBAdapterPostgreSQL.cpp
```diff
@@ -1080,3 +1080,87 @@ int CDBAdapterPostgreSQL::addFilesToDataBase() {
...
   return 0;
 }
+
+int CDBAdapterPostgreSQL::storeLayerMetadata(const char *datasetName, const char *layerName, const char *metadataKey, const char *metadataBlob) {
+#ifdef MEASURETIME
+  StopWatch_Stop(">CDBAdapterPostgreSQL::storeLayerMetadata");
+#endif
+  CPGSQLDB *dataBaseConnection = getDataBaseConnection();
+  if (dataBaseConnection == NULL) {
+    return -1;
```
belentorrente, line 1095: Maybe we could do `CDBError("something something"); throw(__LINE__);` here as well? For debugging purposes in the future.

hclasses/CTString.cpp
```diff
@@ -469,6 +469,25 @@ namespace CT {
     substringSelf(s, e + 1);
   }
 
+  void string::trimWhiteSpacesAndLinesSelf() {
+
+    int s = -1, e = privatelength;
+    const char *value = useStack ? stackValue : heapValue;
+    for (size_t j = 0; j < privatelength; j++) {
+      if (value[j] != ' ' || value[j] != 10 || value[j] != 13) {
```
belentorrente: This should be `&&` instead of `||`? And it is not possible to use the char representation '\n' and '\r' for clarity?

tests/expectedoutputs/TestWMS/test_WMSCMDUpdateDBPathFileInSubfoldersGetCapabilities.xml
```diff
@@ -104,23 +104,23 @@
...
-<BoundingBox CRS="CRS:84" minx="48.895303" miny="0.000000" maxx="55.973600" maxy="10.856452" />
-<BoundingBox CRS="EPSG:25831" minx="282182.345906" miny="5433247.394257" maxx="997135.658676" maxy="6207204.592734" />
-<BoundingBox CRS="EPSG:25832" minx="-153083.018869" miny="5415817.312926" maxx="617595.626092" maxy="6239769.309880" />
-<BoundingBox CRS="EPSG:28992" minx="-236275.338083" miny="106727.731651" maxx="501527.918656" maxy="900797.079725" />
-<BoundingBox CRS="EPSG:7399" minx="0.000000" miny="5757301.056717" maxx="763611.971696" maxy="6483919.801602" />
 <BoundingBox CRS="EPSG:50001" minx="-2000000.000000" miny="-2000000.000000" maxx="10000000.000000" maxy="8500000.000000" />
 <BoundingBox CRS="EPSG:54030" minx="0.000000" miny="5212368.744545" maxx="853649.694601" maxy="5936460.867117" />
-<BoundingBox CRS="EPSG:32661" minx="2000000.000000" miny="-2703305.597319" maxx="2745713.040381" maxy="-1888346.216710" />
-<BoundingBox CRS="EPSG:40000" minx="0.000000" miny="-4731695.771951" maxx="750214.326339" maxy="-3911817.119426" />
+<BoundingBox CRS="EPSG:7399" minx="0.000000" miny="5757301.056717" maxx="763611.971696" maxy="6483919.801602" />
```
belentorrente, line 122: Are these simply formatting changes? Are they related to the new way of generating the XML?

belentorrente (commented):
I just had some minor comments

hclasses/CTString.cpp
maartenplieger: Yes you are right. I changed and now I get what I wanted. If you you place spaces or newlines in the configuration it does now not cause issues.

adagucserverEC/CDBAdapterPostgreSQL.cpp
maartenplieger: Added

adagucserverEC/CDBAdapterPostgreSQL.cpp
maartenplieger: Good point. Lets keep it in mind. Until now it did not cause issues as the -1 one is captured at a later stage.

tests/expectedoutputs/TestWMS/test_WMSCMDUpdateDBPathFileInSubfoldersGetCapabilities.xml
maartenplieger: The ordering has become different and the list has now unique CRS's.

adagucserverEC/CDBAdapterPostgreSQL.cpp
belentorrente: Ah, perfect!

python/python_fastapi_server/routers/edr_utils.py
```diff
@@ -562,8 +562,8 @@ def get_params_for_collection(
             )
         else:
             param_metadata = get_param_metadata(param_id, edr_collection)
-            logger.info(param_id)
-            logger.info(param_metadata["wms_layer_name"])
+            # logger.info(param_id)
+            # logger.info(param_metadata["wms_layer_name"])
```
mgrunbauer: Remove comments?

adagucserverEC/Types/LayerMetadataType.h
```diff
@@ -0,0 +1,66 @@
...
+  int isQueryable = 0;
+  CT::string name, title, group, abstract, nativeEPSG;
+
+  std::vector<LayerMetadataProjection *> projectionList;
+  std::vector<LayerMetadataDim *> dimList;
+  std::vector<LayerMetadataStyle *> styleList;
+  std::vector<LayerMetadataVariable *> variableList;
+};
+
+// TODO should rename this class
```
mgrunbauer, line 56: Do you want to rename it in this MR?

`AdagucLayer`?
`MetadataLayer`?

tests/expectedoutputs/TestWMS/test_WMSGetCapabilities_KMDS_PointNetCDF_pointstylepoint.xml
```diff
@@ -135,51 +185,51 @@
...
+<BoundingBox CRS="EPSG:28992" minx="-8805606.109966" miny="-4164572.069133" maxx="372143.463211" maxy="3233625.747402" />
+<BoundingBox CRS="EPSG:3067" minx="-13588402.263697" miny="1423498.626368" maxx="-747816.881762" maxy="12453122.100930" />
+<BoundingBox CRS="EPSG:32661" minx="-7498001.230434" miny="-8223848.076474" maxx="3272454.133168" maxy="535002.459267" />
+<BoundingBox CRS="EPSG:3411" minx="-3942117.046146" miny="-9975996.132113" maxx="7877215.337643" maxy="-2369697.552660" />
 <BoundingBox CRS="EPSG:3412" minx="-36574088.915720" miny="5641285.215970" maxx="4899857.294031" maxy="39369117.726749" />
 <BoundingBox CRS="EPSG:3575" minx="-7844965.759495" miny="-8002206.437441" maxx="-189084.750364" maxy="-772564.105215" />
 <BoundingBox CRS="EPSG:3857" minx="-7600430.977505" miny="1360506.839359" maxx="795858.888223" maxy="7439724.721958" />
+<BoundingBox CRS="EPSG:40000" minx="-9555333.229813" miny="-10285561.445145" maxx="1280134.942825" maxy="-1473840.584239" />
 <BoundingBox CRS="EPSG:4258" minx="12.130000" miny="-68.275833" maxx="55.399166" maxy="7.149322" />
 <BoundingBox CRS="EPSG:4326" minx="12.130000" miny="-68.275833" maxx="55.399166" maxy="7.149322" />
```
mgrunbauer, line 200: The CRS's get sorted alphabetically, would it be good idea to sort numerically? That way `CRS:3067` would come before `CRS:25831`.

Not sure if it's an advantage for a client who has to parse this, current solution works fine as well.

adagucserverEC/CServerConfig_CPPXSD.h
```diff
@@ -1300,10 +1309,13 @@ class CServerConfig : public CXMLSerializerInterface {
   public:
     class Cattr {
     public:
-      CT::string enablecleanupsystem, cleanupsystemlimit, cache_age_cacheableresources, cache_age_volatileresources;
+      CT::string enablemetadatacache, enablecleanupsystem, cleanupsystemlimit, cache_age_cacheableresources, cache_age_volatileresources;
     } attr;
     void addAttribute(const char *attrname, const char *attrvalue) {
-      if (equals("enablecleanupsystem", attrname)) {
+      if (equals("enablemetadatacache", attrname)) {
+        attr.enablemetadatacache.copy(attrvalue);
```
mgrunbauer, line 1316: `enablemetadatacache` is a new configuration option, I'm guessing you set it like this?:

```xml
  <Settings enablecleanupsystem="{ADAGUCENV_ENABLECLEANUP}" cleanupsystemlimit="1000" cache_age_volatileresources="60" cache_age_cacheableresources="36000" enablemetadatacache="true" />
```

Could this new option be added to the docs? (not sure if we have docs for these options)

adagucserverEC/CDBAdapterPostgreSQL.cpp
```diff
@@ -1080,3 +1080,87 @@ int CDBAdapterPostgreSQL::addFilesToDataBase() {
...
+
+  if (this->layerMetaDataStore == nullptr) {
+#ifdef CDBAdapterPostgreSQL_DEBUG
+    CDBDebug("Need to query layer metadata for %s/%s/%s", datasetName, layerName, metadataKey);
+#endif
+    CPGSQLDB *dataBaseConnection = getDataBaseConnection();
+    if (dataBaseConnection == NULL) {
+      CDBError("No database connection");
+      throw(__LINE__);
+    }
```
mgrunbauer: Maybe not for now, however...

The method `getDataBaseConnection` can return a `NULL` if no connection can be made. In the methods you added we CDBError and throw.
The pre-existing methods immediately return `NULL` without logging or throwing.

Could we move the throw+log inside the `getDataBaseConnection` method? That way we don't have to do the extra `if (conn == NULL)` and get consistent error messages? Or is it a bad idea to always throw inside that method?

adagucserverEC/CDBAdapterPostgreSQL.h
```diff
@@ -43,6 +43,8 @@ class CDBAdapterPostgreSQL : public CDBAdapter {
   std::map<std::string, std::vector<std::string>> fileListPerTable;
   int createDimTableOfType(const char *dimname, const char *tablename, int type);
 
+  CDBStore::Store *layerMetaDataStore = nullptr;
```
mgrunbauer, line 47: It seems we don't `delete this->layerMetaDataStore` anywhere.

Should we delete this somewhere after we're done using it?

adagucserverEC/CDBAdapterPostgreSQL.cpp
```diff
@@ -1080,3 +1080,87 @@ int CDBAdapterPostgreSQL::addFilesToDataBase() {
...
+    this->layerMetaDataStore = dataBaseConnection->queryToStore(query.c_str());
+    if (layerMetaDataStore == nullptr) {
+#ifdef CDBAdapterPostgreSQL_DEBUG
+      CDBDebug("Unable query: \"%s\"", query.c_str());
+#endif
+      throw(__LINE__);
+    }
+    if (layerMetaDataStore->size() == 0) {
+      CDBDebug("No results \"%s\"", query.c_str());
+      throw(__LINE__);
```
mgrunbauer: If `layerMetaDataStore->size() == 0`, it means the query did not result entries. This would happen on a clean adaguc deploy I think.

If we `throw` here, do we try to create the `GetCapabilities` document the regular way?

adagucserverEC/utils/LayerMetadataStore.cpp
```diff
@@ -0,0 +1,307 @@
...
+}
+
+// Not sure if this will be used
+int loadMyWMSLayerFromMetadataDb(WMSLayer *myWMSLayer) {
+  loadLayerMetadataStructFromMetadataDb(myWMSLayer);
+  loadLayerDimensionListFromMetadataDb(myWMSLayer);
+  loadLayerProjectionAndExtentListFromMetadataDb(myWMSLayer);
+  loadLayerStyleListFromMetadataDb(myWMSLayer);
+  return 0;
+}
```
mgrunbauer: Used for testing?

adagucserverEC/CDBFileScanner.cpp
```diff
@@ -820,7 +821,7 @@ int CDBFileScanner::DBLoopFiles(CDataSource *dataSource, int removeNonExistingFi
               // delete cdfObject;cdfObject=NULL;
               // cdfObject=CDFObjectStore::getCDFObjectStore()->deleteCDFObject(&cdfObject);
             } catch (int linenr) {
-              CDBError("Exception in DBLoopFiles at line %d");
+              CDBError("Exception in DBLoopFiles at line %d", linenr);
```
mgrunbauer, line 825: I thought `CDBError` already prints `linenr`, or does it work different since this is the scanner and not the `adagucserver` that we spawn?

from `CDebugger.h`:
```
#define CDBError                                                                                                                                                                                       \
  _printError("[E:%03d:pid%lu: %s:%d] ", logMessageNumber, logProcessIdentifier, __FILENAME__, __LINE__);                                                                                              \
  _printErrorLine
```

adagucserverEC/CDBAdapterPostgreSQL.cpp
maartenplieger: I removed the logging here. The throw is need in this case, because of the return type which is CT::string

adagucserverEC/CDBAdapterPostgreSQL.h
maartenplieger: Added a delete. Good point!

adagucserverEC/CDBAdapterPostgreSQL.cpp
maartenplieger: I made it gracefully fail, by returning "". Tested it by deleting the layermeratastore, it just gets slower, nothing else.

adagucserverEC/utils/LayerMetadataStore.cpp
maartenplieger: Removed it.

adagucserverEC/CDBFileScanner.cpp
maartenplieger: It says at which line the exception was thrown. So you will now where the error is logged, and which line the exception is thrown.

python/python_fastapi_server/routers/edr_utils.py
maartenplieger: Removed

adagucserverEC/Types/LayerMetadataType.h
maartenplieger: Renamed to MetadataLayer

tests/expectedoutputs/TestWMS/test_WMSGetCapabilities_KMDS_PointNetCDF_pointstylepoint.xml
maartenplieger: I would like to leave as is, I think it does not cause any issues and it is consistent like this.

adagucserverEC/CServerConfig_CPPXSD.h
maartenplieger: Added documentation

tests/expectedoutputs/TestWMS/test_WMSGetCapabilities_KMDS_PointNetCDF_pointstylepoint.xml
mgrunbauer: Ok, super! Note that this MR will help fix: https://github.com/KNMI/adaguc-server/issues/350

adagucserverEC/CDBFileScanner.cpp
mgrunbauer: Ah, I understand, clear!

CCDFDataModel/CCDFVariable.h
```diff
@@ -291,21 +291,29 @@ namespace CDF {
     void setSize(size_t size) { currentSize = size; }
     size_t getSize() { return currentSize; }
 
-    Attribute *getAttribute(const char *name) const {
+    bool hasAttribute(const char *name) const {
```
lukas-phaf: Function doesn't seem to be used.

CCDFDataModel/CCDFVariable.h
```diff
@@ -291,21 +291,29 @@ namespace CDF {
...
+    }
+
+    Attribute *getAttribute(const char *name) const {
+      Attribute *a = getAttributeNE(name);
+      if (a == nullptr) {
+        throw(CDF_E_ATTNOTFOUND);
+      }
+      return a;
     }
     Attribute *getAttributeNE(const char *name) const {
```
lukas-phaf, line 302: Add newline.

CCDFDataModel/CCDFVariable.h
```diff
@@ -291,21 +291,29 @@ namespace CDF {
...
-      try {
-        return getAttribute(name);
-      } catch (int e) {
-        return NULL;
+      for (size_t j = 0; j < attributes.size(); j++) {
+        if (attributes[j]->name.equals(name)) {
```
lukas-phaf: CLion suggests this, but you might not want to make this change now:

```suggestion
      for (auto attribute : attributes) {
        if (attribute->name.equals(name)) {
```

NEWS.md
```diff
@@ -1,3 +1,7 @@
+**Version 2.28.0 2024-09-62**
```
lukas-phaf: When's that?  ;-)

NEWS.md
```diff
@@ -1,3 +1,7 @@
...
+- Metadata for layer items like variables, projections, dimensions and styles are now stored in a metadatatable. This can be disabled via the `enablemetadatacache` property in [Settings](doc/configuration/Settings.md).
```
lukas-phaf:
```suggestion
- Metadata for layer items like variables, projections, dimensions and styles are now stored in a database table called `layermetadata`. This can be disabled via the `enablemetadatacache` property in [Settings](doc/configuration/Settings.md).
```

adagucserverEC/CAutoConfigure.cpp
```diff
@@ -484,10 +484,11 @@ int CAutoConfigure::justLoadAFileHeader(CDataSource *dataSource) {
...
 
-  CT::string foundFileName;
-
-  /* Use the file specified as header file */
-  foundFileName = dataSource->headerFileName.c_str();
+  CT::string foundFileName = dataSource->getFileName();
+  if (foundFileName.empty()) {
+    /* Use the file specified as header file */
+    foundFileName = dataSource->headerFileName.c_str();
+  }
```
lukas-phaf: It is not clear to me what the difference between `fileName` and `headerFileName` is, and how this is affected by animationStep.

adagucserverEC/CDBAdapterSQLLite.h
```diff
@@ -119,5 +119,7 @@ class CDBAdapterSQLLite : public CDBAdapter {
   int setFileString(const char *tablename, const char *file, const char *dimvalue, int dimindex, const char *filedate, GeoOptions *geoOptions);
   int setFileTimeStamp(const char *tablename, const char *file, const char *dimvalue, int dimindex, const char *filedate, GeoOptions *geoOptions);
   int addFilesToDataBase();
+  int storeLayerMetadata(const char *datasetName, const char *layerName, const char *metadataKey, const char *metadatablob);
+  CT::string getLayerMetadata(const char *datasetName, const char *layerName, const char *metadataKey);
```
lukas-phaf: What happens if "store metadata" is turned on when SQLLite is in use?

adagucserverEC/CDBAdapterPostgreSQL.cpp
```diff
@@ -1080,3 +1084,86 @@ int CDBAdapterPostgreSQL::addFilesToDataBase() {
...
+int CDBAdapterPostgreSQL::storeLayerMetadata(const char *datasetName, const char *layerName, const char *metadataKey, const char *metadataBlob) {
+#ifdef MEASURETIME
+  StopWatch_Stop(">CDBAdapterPostgreSQL::storeLayerMetadata");
+#endif
+  CPGSQLDB *dataBaseConnection = getDataBaseConnection();
+  if (dataBaseConnection == NULL) {
+    return -1;
+  }
+
+  CT::string tableColumns("datasetname varchar (255), layername varchar (255), metadatakey varchar (255), updatetime TIMESTAMPTZ, blob JSONB, PRIMARY KEY (datasetname, layername, metadatakey)");
```
lukas-phaf: Two things:
- We use a mix of 255 and 511 length voor `varchar`. Why did you pick 255 here?
- In other places I see `UNIQUE`, not `PRIMARY KEY`. Looking at the [docs](https://www.postgresql.org/docs/current/sql-createtable.html), I think `PRIMARY KEY` is better here. But not clear why it is not used in other places. But not a big deal.

adagucserverEC/CDBAdapterPostgreSQL.cpp
```diff
@@ -1080,3 +1084,86 @@ int CDBAdapterPostgreSQL::addFilesToDataBase() {
...
+
+  int status = dataBaseConnection->checkTable("layermetadata", tableColumns.c_str());
+  if (status == 1) {
+    CDBError("\nFAIL: Table layermetadata could not be created: %s", tableColumns.c_str());
+    throw(__LINE__);
+  }
+
+  CT::string updateTime = CTime::currentDateTime().c_str();
+  updateTime.setSize(19);
```
lukas-phaf: What does this do? Convert through c_str(), and resive?

adagucserverEC/CDBAdapterPostgreSQL.cpp
```diff
@@ -1080,3 +1084,86 @@ int CDBAdapterPostgreSQL::addFilesToDataBase() {
...
+  CT::string query;
+  query.print("INSERT INTO layermetadata values (E'%s',E'%s', E'%s', E'%s', E'%s') "
```
lukas-phaf: It is best practice to explicitly list the column values when inserting, incase somebody adds columns or changes the order.

If you want to keep it like it is, please capitalize `values`.

adagucserverEC/CDBAdapterPostgreSQL.cpp
```diff
@@ -1080,3 +1084,86 @@ int CDBAdapterPostgreSQL::addFilesToDataBase() {
...
+#ifdef CDBAdapterPostgreSQL_DEBUG
+    CDBDebug("Need to query layer metadata for %s/%s/%s", datasetName, layerName, metadataKey);
+#endif
+    CPGSQLDB *dataBaseConnection = getDataBaseConnection();
+    if (dataBaseConnection == NULL) {
+      throw(__LINE__);
+    }
+
+    CT::string query;
+    query.print("SELECT layername, metadatakey, blob from layermetadata where datasetname = '%s';", datasetName);
```
lukas-phaf: I know we haven't so far worried about SQL injection, but this one seems a bit easy to access:
Do a GetCapabilties call with a `DROP TABLES` some where in the dataset name would do the trick, I think?

Probably for the already existing DB tables, the queries are by layer name, and these are checked in the dataset configs?

adagucserverEC/CAutoConfigure.cpp
maartenplieger: - `dataSource->getFileName()` has the filename for the current requested set of dimension. This is used when reading actual data. But it is sometimes not yet present, as in certain cases the header information needs to be read first.
- `dataSource->headerFileName()` points to one of the files in the sequence of files for a layer. This can be used to read metadata.

justLoadAFileHeader tries to load netcdf header information. When there is already a file name available in headerFileName, that can be used. It will prevent costly directory scans or database lookups.

adagucserverEC/CDBAdapterSQLLite.h
maartenplieger: Nothing. 
You can test it via 
`bash runtests.sh`
`bash runtests_psql.sh`

It is backwards compatible and it will provide the same results.

adagucserverEC/CDBAdapterPostgreSQL.cpp
```diff
@@ -1080,3 +1084,86 @@ int CDBAdapterPostgreSQL::addFilesToDataBase() {
...
+#endif
+  return 0;
+}
+
+CT::string CDBAdapterPostgreSQL::getLayerMetadata(const char *datasetName, const char *layerName, const char *metadataKey) {
+#ifdef MEASURETIME
+  StopWatch_Stop(">CDBAdapterPostgreSQL::getLayerMetadata");
+#endif
+
+  if (this->layerMetaDataStore == nullptr) {
```
lukas-phaf, line 1130: This is a cache, right? The idea is to avoid multiple queries to the DB when this function gets called for different layers and `metadataKey`, right?

What if this function get called again with a different `datasetName`?

CCDFDataModel/CCDFVariable.h
maartenplieger: Removed

CCDFDataModel/CCDFVariable.h
maartenplieger: Done

CCDFDataModel/CCDFVariable.h
maartenplieger: We should refactor this class altogether in another PR. Also the throw stuff is costly and could be done with a return statement.

NEWS.md
maartenplieger: Ahahaha, Well spotted.

NEWS.md
maartenplieger: Good suggestion, changed

adagucserverEC/CDBAdapterPostgreSQL.cpp
lukas-phaf:
```suggestion
    query.print("SELECT layername, metadatakey, blob FROM layermetadata WHERE datasetname = '%s';", datasetName);
```

adagucserverEC/CDBAdapterPostgreSQL.h
lukas-phaf, line 47: This file has no `#ifndef` guards. Same for `CDBAdapterSQLLite.h`. They feel left out.

adagucserverEC/CDBAdapterPostgreSQL.cpp
```diff
@@ -1080,3 +1084,86 @@ int CDBAdapterPostgreSQL::addFilesToDataBase() {
 #endif
   return 0;
 }
+
+int CDBAdapterPostgreSQL::storeLayerMetadata(const char *datasetName, const char *layerName, const char *metadataKey, const char *metadataBlob) {
```
lukas-phaf, line 1089: Just so I understand correctly:
The table doesn't have seperate columns for the different metadata bits, but instead they are stored in seperate rows, identified by `metadataKey`?

These rows are stored one by one, but always queried together, right?

adagucserverEC/CDBAdapterPostgreSQL.cpp
maartenplieger: Yes it is cache, in this case for this dataset.

There are currently no cases for multiple datasets in one process.

adagucserverEC/CDBAdapterPostgreSQL.cpp
maartenplieger: Yes

adagucserverEC/CDBAdapterPostgreSQL.cpp
maartenplieger: Lets keep `PRIMARY KEY` here. I would avoid changing the others.



Changed the varchar to 511 for datasetname and layername, although 255 is probably plenty.

adagucserverEC/CDBAdapterPostgreSQL.cpp
maartenplieger: Had to reduce the size because the postgresl timestamp column does not accept the trailing `Z`

adagucserverEC/CDBAdapterPostgreSQL.cpp
maartenplieger: Changed with your suggestion. I like it.

adagucserverEC/CDBAdapterPostgreSQL.cpp
maartenplieger: Yes, the datasetname is checked early in the process. 

Nevertheless I added the same check, just in case we forget  or misuse it in another way.

maartenplieger: Done

adagucserverEC/CDBAdapterPostgreSQL.h
maartenplieger: I hope that they are happy with their guards now ;)

adagucserverEC/CDBAdapterPostgreSQL.cpp
lukas-phaf: Ok. Another thing is that `CTime::currentDateTime()` already returns a CT::string, why go through `c_str()`?

lukas-phaf: Did you check if the times in the table are correct if the DB and client are not running in UTC time zone?
Because I think inserting a time without timeonze into time `TIMESTAMPTZ` column might trigger timezone conversion.

See [here](https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-DATETIME-INPUT-TIME-STAMPS):
`For timestamp with time zone, the internally stored value is always in UTC (Universal Coordinated Time, traditionally known as Greenwich Mean Time, GMT). An input value that has an explicit time zone specified is converted to UTC using the appropriate offset for that time zone. If no time zone is stated in the input string, then it is assumed to be in the time zone indicated by the system's [TimeZone](https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-TIMEZONE) parameter, and is converted to UTC using the offset for the timezone zone.`

Can we get the timestamp in a format like this to avoid the issue:
`1999-01-08 04:05:06 -8:00`

Maybe instead of resizing, just replace `Z` by ` 0:00`?

adagucserverEC/CDBAdapterPostgreSQL.cpp
lukas-phaf: That is true (as all the tests works...), but at the very least the function comments should make this very clear. Or be safe and implement the case by storing the values in a map by`datasetName`.

adagucserverEC/CDBAdapterPostgreSQL.cpp
lukas-phaf: `checkIfPathHasValidTokens()` is in other cases use to validate path, not sure it helps against SQL injection.

I'll look at the other check.

adagucserverEC/CDBAdapterPostgreSQL.cpp
lukas-phaf: Is there a particular reason for this? Does it make the logic/code easier?

adagucserverEC/utils/LayerMetadataStore.cpp
```diff
@@ -0,0 +1,298 @@
...
+  layerMetadataItem["cellsizey"] = myWMSLayer->layerMetadata.cellsizeY;
+  json variables;
+  for (auto lv : myWMSLayer->layerMetadata.variableList) {
+    json variable;
+    variable["units"] = lv->units;
+    variables.push_back(variable);
+  }
+  layerMetadataItem["variables"] = variables;
+
+  storeLayerMetadataInDb(myWMSLayer, "layermetadata", layerMetadataItem.dump());
```
lukas-phaf: The other 3 store functions have `try catch`, but this one does not.

adagucserverEC/utils/LayerMetadataStore.cpp
```diff
@@ -0,0 +1,298 @@
...
+  if (datasetName.empty()) {
+    // CDBDebug("Not a dataset");
+    return "";
+  }
+  return CDBFactory::getDBAdapter(myWMSLayer->dataSource->srvParams->cfg)->getLayerMetadata(datasetName, layerName, metadataKey);
+}
+
+int storeLayerMetadataInDb(MetadataLayer *myWMSLayer, CT::string metadataKey, std::string metadataBlob) {
+  try {
+    CT::string datasetName = myWMSLayer->dataSource->srvParams->datasetLocation;
```
lukas-phaf: `srvParams` seems like a strange place to get/store the dataset name?

adagucserverEC/Types/LayerMetadataType.h
```diff
@@ -0,0 +1,66 @@
...
+  double cellsizeY = 0;
+  double dfLatLonBBOX[4] = {-180, -90, 180, 90};
+  double dfBBOX[4] = {-180, -90, 180, 90};
+  int isQueryable = 0;
+  CT::string name, title, group, abstract, nativeEPSG;
+
+  std::vector<LayerMetadataProjection *> projectionList;
+  std::vector<LayerMetadataDim *> dimList;
+  std::vector<LayerMetadataStyle *> styleList;
+  std::vector<LayerMetadataVariable *> variableList;
```
lukas-phaf: Why use pointers here, instead of putting the structs directly into the vectors?
Construct the structs with something like 
`LayerMetadataStyle sList = {name, title, abstact}`
followed by `push_back`.

If you are worried about performance:
1. Don't be, it won't matter.
2. Copy elision and move semantics probably remove all the copies anyway.
3. If you are still worried, provide a move constructor on the structs and use `emplace_back()`. 

And example of point 3 (but you don't need to go there...):
```
#include <iostream>
#include <vector>

struct style {
  std::string name;
  std::string title;
  std::string abstract;
  
  style(std::string name, std::string title, std::string abstract)
    : name(std::move(name)), title(std::move(title)), abstract(std::move(abstract))
    {}
};

int main() {
    style s = {"test32", "test31", "test30"};
    v.push_back(s);

    v.emplace_back("test34", "test33", "test99");
    
    std::cout<<v[0].name<<std::endl;
    std::cout<<v[1].abstract<<std::endl;
}
```
According to the [spec](https://en.cppreference.com/w/cpp/container/vector/emplace_back) this avoids all copies.

adagucserverEC/utils/XMLGenUtils.cpp
```diff
@@ -0,0 +1,836 @@
...
+  delete styleListFromDataSource;
+
+  return 0;
+}
+
+bool compareStringCase(const std::string &s1, const std::string &s2) { return strcmp(s1.c_str(), s2.c_str()) <= 0; }
+
+bool compareProjection(const LayerMetadataProjection *p1, const LayerMetadataProjection *p2) { return strcmp(p1->name.c_str(), p2->name.c_str()) <= 0; }
+bool compareDim(const LayerMetadataDim *p2, const LayerMetadataDim *p1) { return strcmp(p1->name.c_str(), p2->name.c_str()) <= 0; }
+bool compareStyle(const LayerMetadataStyle *p1, const LayerMetadataStyle *p2) { return strcmp(p2->name.c_str(), p1->name.c_str()) <= 0; }
```
lukas-phaf: Function not used.

adagucserverEC/utils/XMLGenUtils.h
```diff
@@ -0,0 +1,16 @@
...
+#include "CXMLGen.h"
+
+int getProjectionInformationForLayer(MetadataLayer *myWMSLayer);
+int getDimsForLayer(MetadataLayer *myWMSLayer);
+int getStylesForLayer(MetadataLayer *myWMSLayer);
+bool compareStringCase(const std::string &s1, const std::string &s2);
+bool compareProjection(const LayerMetadataProjection *p1, const LayerMetadataProjection *p2);
+bool compareDim(const LayerMetadataDim *p1, const LayerMetadataDim *p2);
+bool compareStyle(const LayerMetadataStyle *p1, const LayerMetadataStyle *p2);
+int populateMyWMSLayerStruct(MetadataLayer *myWMSLayer, bool readFromDb);
```
lukas-phaf: Clean up the names?

adagucserverEC/CDataPostProcessors/CDataPostProcessor_Operator.cpp
```diff
@@ -64,8 +64,8 @@ int CDPPOperator::execute(CServerConfig::XMLE_DataPostProc *proc, CDataSource *d
       newDataObject->cdfVariable->setAttributeText("units", proc->attr.units);
     }
 
-    short attrData[3];
-    attrData[0] = -1;
+    double attrData[1];
+    attrData[0] = 0;
```
lukas-phaf, line 68: Was this change intended?

adagucserverEC/CDataSource.h
```diff
@@ -373,6 +373,8 @@ class CDataSource {
    * @return 0 on succes, 1 on failure
    */
   int readVariableDataForCDFDims(CDF::Variable *variableToRead, CDFType dataTypeToReturnData);
+
+  std::string getDataSetName();
```
lukas-phaf, line 377: Use this in `LayerMetadataStore`? It is now not used.

adagucserverEC/CDBAdapterPostgreSQL.cpp
maartenplieger: I was not able to insert timestamps with the suggested format. But I changed the type from TIMESTAMPTZ to TIMESTAMP and now I see UTC timestamps in the table.

adagucserverEC/CDBAdapterPostgreSQL.cpp
maartenplieger: We use 
```
CServerParams::checkForValidTokens(datasetName, "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz_-:/."
```

adagucserverEC/CDBAdapterPostgreSQL.cpp
maartenplieger: That will be done in the follow up MR where we want to return all datasets with their layer info using one request.

adagucserverEC/CDBAdapterPostgreSQL.cpp
maartenplieger: Yes, metadata for the layer struct is collected at different places and then immediately stored in the table. 

This gives the opportunity to only update the dimensions and leave the projection, styles and layermetadata alone.

adagucserverEC/utils/LayerMetadataStore.cpp
maartenplieger: srvParams is populated with items from the query string when the process runs. It is passed around to obtain information like format, datasetname, layers, etc....

hclasses/CTString.cpp
```diff
@@ -469,6 +469,25 @@ namespace CT {
...
+    }
+    for (size_t j = privatelength - 1; j > 0; j--) {
+      if (value[j] != ' ' && value[j] != '\n' && value[j] != '\r') {
+        e = j;
+        break;
+      }
+    }
+    substringSelf(s, e + 1);
+  }
+
```
lukas-phaf: It hurts that we are writing stuff like this ourselves. Or copy & pasting it...

Anyway, can we merge it with the above, give that a `bool` flag that defaults to `false` that lets it also remove newline stuff?

adagucserverEC/utils/LayerMetadataStore.cpp
maartenplieger: storeLayerMetadataInDb does not throw anymore

adagucserverEC/Types/LayerMetadataType.h
maartenplieger: Changed from pointers to directly using structs

adagucserverEC/utils/XMLGenUtils.cpp
maartenplieger: Removed

adagucserverEC/utils/XMLGenUtils.h
maartenplieger: Renamed

adagucserverEC/CDataPostProcessors/CDataPostProcessor_Operator.cpp
maartenplieger: Yes, the original fillvalue was wrong, I saw this in the tests.

adagucserverEC/CDataSource.h
maartenplieger: It is now used

hclasses/CTString.cpp
maartenplieger: Moved into the same function with an optional argument.

ernstdevreede (approved):

PR #410, opened 2024-10-09
Windspeed barbs now display text in kts next to the barb

maartenplieger:
**Version 2.28.1 2024-10-11**

- Windbarbs on modelfields can now display the windspeed in kts as text when rendertextforvectors in RenderSettings is set to true.


tests/expectedoutputs/TestCSV/test_CSV_windbarbs.png
lukas-phaf, line 1: It could already show the speed in KTS next to it, so what exactly is new in this PR?

lukas-phaf, line 1: Why have 0, 1 and 2 changed? Is the fontsize for 0,1 and 2 now different, while it was the same before?

tests/expectedoutputs/TestWMS/test_WMSGetMapWithHarmWindBarbs_without_outline.png
lukas-phaf, line 1: Gird has changed, is that expected?

adagucserverEC/CCairoPlotter.cpp
```diff
@@ -1252,6 +1252,8 @@ void CCairoPlotter::drawBarb(int x, int y, double direction, double strength, CC
   CT::string text;
   text.print("%d", strengthInKnots);
   if (drawText) {
-    this->drawStrokedText(x - cos(direction + M_PI) * 15 - 5, y + sin(direction + M_PI) * 12 + 5, 0, text.c_str(), 12, 1 * drawOutline, outlineColor, barbColor);
+    // If speed is really low, draw the text below the circle
+    double textDirection = strengthInKnotsRoundedToFive <= 2 ? -M_PI / 2.1 : direction;
+    this->drawStrokedText(x - cos(textDirection + M_PI) * 15 - 5, y + sin(textDirection + M_PI) * 12 + 5, 0, text.c_str(), 12, 1 * drawOutline, outlineColor, barbColor);
```
belentorrente, line 1257: Maybe this is something out of the scope of the PR but good for the future refactoring, but I see a lot of "magic numbers" and it is difficult to follow their meaning (for example, why 15? why 2.1).

maartenplieger: Yes, this is for text placement of the number so it visually appears at the right place.
Since it is captured in tests, and since we want to refactor I would leave it like this.

tests/expectedoutputs/TestCSV/test_CSV_windbarbs.png
maartenplieger: I think this is an optical illusion. I checked with gimp, the font size is the same.

image

belentorrente (approved):
Looks good to me

PR #411, opened 2024-10-17
Refactored scan scripts so they work with non standard directories

maartenplieger:
- Re-organized scanning scripts
- Backwards compatible with old methods

Old method:
```
docker exec -i -t my-adaguc-server /adaguc/adaguc-server-updatedatasets.sh
```

New method:

```
docker exec -i -t my-adaguc-server /adaguc/scan.sh
```

Or local without docker:
```
./scripts/scan.sh
```

The scan script can:
- list files with `docker exec -i -t my-adaguc-server /adaguc/scan.sh -l`
- list environment `docker exec -i -t my-adaguc-server /adaguc/scan.sh -e`
- scan a dataset `docker exec -i -t my-adaguc-server /adaguc/scan.sh -d <datasetname>`
- scan a file  `docker exec -i -t my-adaguc-server /adaguc/scan.sh -f <filename>`
- scan all datasets with `docker exec -i -t my-adaguc-server /adaguc/scan.sh`

The scripts are put in the ./scripts/ folder. They do work in local development and in Docker environment.


Docker/adaguc-server-cleandb.sh
```diff
@@ -1,125 +0,0 @@
-#!/bin/bash
```
mgrunbauer: The file `adaguc-server-cleandb.sh` gets used in the nested stack (`clean_files_from_database`).

Should we keep this file to support cleaning the database, or do you plan on removing support for this completely?

Suggestion: what about a `clean.sh` that supports cleaning the database, datasets (xml) and data (nc) from disk?

Docker/adaguc-server-addfile.sh
```diff
@@ -1,90 +1,2 @@
...
-  command="${ADAGUC_PATH}/bin/adagucserver --updatedb --verboseoff --config ${ADAGUC_CONFIG},${ADAGUC_DATASET}"
-  echo $command
-  $command
-  OUT=$?
-  if [ ${OUT} -ne 0 ]; then
-    STATUSCODE=1
-  fi
-  exit ${STATUSCODE} 
-fi
+. ${ADAGUC_PATH}/scripts/scan.sh
```
mgrunbauer: This will scan _all_ datasets, the previous behaviour was to scan only the given dataset with `-d` or the given file with `-f`.

Note that this also gets used in the nested stack.

What about adding `"$@"` at the end to pass all the command line parameters to `scan.sh`? This way, the following commands still work as before:

```bash
adaguc-server-addfile.sh -f myfile.nc
adaguc-server-addfile.sh -d mydataset
```

scripts/scan.sh
```diff
@@ -0,0 +1,105 @@
...
+  echo $command
+  $command
+  OUT=$?
+  if [ ${OUT} -ne 0 ]; then
+    STATUSCODE=1
+  fi
+  exit ${STATUSCODE} 
+fi
+
+# Scan al datasets
```
mgrunbauer: Typo `al` -> `all`

scripts/scan.sh
```diff
@@ -0,0 +1,105 @@
...
+    echo "Scanning full dataset [${configfile}]:"
+    command="${ADAGUC_PATH}/bin/adagucserver --updatedb --verboseoff --config ${ADAGUC_CONFIG},${configfile}"
+    echo $command
+    $command
+    OUT=$?
+    if [ ${OUT} -ne 0 ]; then
+      STATUSCODE=1
+    fi
+  done
+  exit ${STATUSCODE}
```
mgrunbauer: If `$command` exit code does not equal `0`, `STATUSCODE` gets set to 1. However we then set `STATUSCODE=0` on the next iteration, clearing a possible failure.

If you scan two datasets, the first one fails but the second succeeds, this script will exit with `0`.

scripts/scan.sh
```diff
@@ -0,0 +1,105 @@
...
+  STATUSCODE=0
+  echo "Adding file [${ADAGUC_DATAFILE}] to dataset [${ADAGUC_DATASET}]:"
+  command="${ADAGUC_PATH}/bin/adagucserver --updatedb --config ${ADAGUC_CONFIG},${ADAGUC_DATASET} --path ${ADAGUC_DATAFILE}"
+  echo $command
+  $command
+  OUT=$?
+  if [ ${OUT} -ne 0 ]; then
+    STATUSCODE=1
+  fi
+  exit ${STATUSCODE}
```
mgrunbauer: Why always set `STATUSCODE=1` on failure, maybe we can exit with the original exit code from `$command`?

scripts/scan.sh
```diff
@@ -0,0 +1,105 @@
...
+
+### Scan a file ###
+if [[ -n "${ADAGUC_DATAFILE}" ]]; then
+
+  basediroffile=${ADAGUC_DATAFILE%/*}
+  basenameoffile="${ADAGUC_DATAFILE##*/}"
+  
+  STATUSCODE=0
+  echo "Adding file [${ADAGUC_DATAFILE}] to dataset [${alldatasets}]:"
+  command="${ADAGUC_PATH}/bin/adagucserver --updatedb --autofinddataset --verboseoff --config ${ADAGUC_CONFIG} --path ${ADAGUC_DATAFILE}"
```
mgrunbauer: Some of the commands set `--verboseoff`, but not all. Not sure how much gets hidden with `--verboseoff`.

Maybe you could add a `-v|--verbose` option to `scan.sh` which controls the `--verboseoff`?

That way you could do something like `scan.sh -v -d mydataset`.

scripts/scan.sh
```diff
@@ -0,0 +1,105 @@
...
+# With -f you can optionally specify the file to add
+# With -d you can optionally specify the dataset to add.
+# If only a file is specified without a dataset, adaguc-server will try to find the matching dataset that belongs to the file.
+#
+# Usage: ./scan.sh -d <datasetname (optional)> -f <file to add>
+# Usage: ./scan.sh -f <file to add> 
+# Usage: ./scan.sh -d <datasetname> 
+# Usage: ./scan.sh -d "*" to scan all datasets
+# Usage: ./scan.sh -l to list datasets
+# Usage: ./scan.sh -e to see current environment
```
mgrunbauer: You could add this to a `usage()` function and print this when you pass `-h` (or an unknown flag)? One of the original scripts also did this.

doc/EnvironmentVariables.md
```diff
@@ -0,0 +1,55 @@
+# Environment variables
```
mgrunbauer, line 1: Nice, I like this! 👍 

At least two environment variables are missing:
```
PGBOUNCER_ENABLE 	Enable or disable the usage of PostgreSQL connection pooling by PGBouncer. Default: true.
PGBOUNCER_DISABLE_SSL 	If PGBouncer is used, disable the usage of SSL. Default: true.
```

Note: I was working on something similar [in my fork branch](https://github.com/KNMI/adaguc-server/blob/7739589cfe721caa6314ed69b69b2a97fae5d0b7/doc/Environment_Variables.md) with a formatted table. It includes a few more environment variables but some might be old/unused/deprecated :)

Docker/adaguc-server-cleandb.sh
maartenplieger: Reverted that file

Docker/adaguc-server-addfile.sh
maartenplieger: Good idea, added

scripts/scan.sh
maartenplieger: Changed

scripts/scan.sh
maartenplieger: Good catch!

scripts/scan.sh
maartenplieger: Now setting to returncode of the process.

scripts/scan.sh
maartenplieger: Added -v option

scripts/scan.sh
maartenplieger: Added!

doc/EnvironmentVariables.md
maartenplieger: Added

mgrunbauer (approved):

PR #413, opened 2024-10-22
Fix bug which caused scheduled metadata updates to only work intermittently

lukas-phaf:


python/lib/adaguc/runAdaguc.py
```diff
@@ -101,16 +101,13 @@ def getAdagucEnv(self, adagucenv = {}):
 
         return adagucenv
 
-    def updateLayerMetadata(self):
+    async def updateLayerMetadata(self):
         """Uses the adaguc executable to update the layermetadatatable"""
         adagucenv = self.getAdagucEnv()
-        status, data, headers = asyncio.run(
```
lukas-phaf, line 107: @maartenplieger, I think this was the problem.

maartenplieger (dismissed):

python/lib/adaguc/runAdaguc.py
```diff
@@ -101,16 +101,13 @@ def getAdagucEnv(self, adagucenv = {}):
 
         return adagucenv
 
-    def updateLayerMetadata(self):
+    async def updateLayerMetadata(self):
```
maartenplieger, line 104: This resolves the issue.

Can you bump the version in Definitions.h, in the Dockerfile and write a short section in News.md?

maartenplieger (approved):

PR #415, opened 2024-10-24
Fix bug where directory reader could not figure out the file type (opendir ent->d_type == DT_UNKNOWN)

maartenplieger:
- Added logging statements to verify what is going on in updatelayer metadata
- Fix bug where directory reader could not figure out the file type (opendir ent->d_type == DT_UNKNOWN)
- Updatelayermetadata is not going to do anything if no datasets or layers are found

### DIFF
--- a/hclasses/CDirReader.cpp
+++ b/hclasses/CDirReader.cpp
@@ -51,7 +51,6 @@ std::vector<std::string> CDirReader::listDir(const char *directory, bool recursi
   DIR *dir;
   struct dirent *ent;
   if ((dir = opendir(directory)) != NULL) {
-    errno = 0;
     /* print all the files and directories within directory */
     while ((ent = readdir(dir)) != NULL) {
       CT::string fullName = directory;
@@ -88,13 +87,6 @@ std::vector<std::string> CDirReader::listDir(const char *directory, bool recursi
       }
     }
     closedir(dir);
-    if (errno != 0) {
-      if (exceptionOnError) {
-        CDBWarning("errno != 0: Unable to readdir [%s]", directory);
-        /* could read directory */
-        throw "Error: Could not readdir";
-      }
-    }
   } else {
     if (exceptionOnError) {
       CDBWarning("Unable to open dir [%s]", directory);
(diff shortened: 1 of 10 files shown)


hclasses/CDirReader.cpp
```diff
@@ -53,18 +53,30 @@ std::vector<std::string> CDirReader::listDir(const char *directory, bool recursi
...
+          d_type = DT_REG;
+        }
+        if (ret == 0 && S_ISDIR(path_stat.st_mode)) {
+          d_type = DT_DIR;
+        }
+      }
 
-          if (((filesAndOrDirs & CDIRREADER_INCLUDE_FILES) && ent->d_type == DT_REG) || ((filesAndOrDirs & CDIRREADER_INCLUDE_DIRECTORIES) && ent->d_type == DT_DIR)) {
+      if (d_type == DT_REG || d_type == DT_DIR) {
+        if (ent->d_name[0] != '.') { // Omit dirs starting with a .
```
mgrunbauer: Note, the comment says "omit dirs starting with a ." but we are also omitting files starting with a ".". The file `.my-dataset.xml` would pass the regex but still get excluded.

Is this intended behavior?
If yes, then maybe update the comment?

hclasses/CDirReader.cpp
```diff
@@ -53,18 +53,30 @@ std::vector<std::string> CDirReader::listDir(const char *directory, bool recursi
   if ((dir = opendir(directory)) != NULL) {
     /* print all the files and directories within directory */
     while ((ent = readdir(dir)) != NULL) {
```
mgrunbauer, line 55: It was already like this, but for reference, `man readdir` says:

> In the event of an error, errno may be set to any of the values documented for the getdirentries(2) system call.

And `man 2 getdirentries` shows the possible errors:

>     [EBADF]            fd is not a valid file descriptor open for reading.
>
>     [EFAULT]           Either buf or basep point outside the allocated address space.
>
>     [EIO]              An I/O error occurred while reading from or writing to the file system.

We do not check `errno` currently. If we want to handle these cases, we should also check for `errno`, but it might be too much for this fix?

mgrunbauer (dismissed):
Approved, left minor suggestions.

hclasses/CDirReader.cpp
maartenplieger: Good point

maartenplieger: Removed it again as it is not clear when what how errno exactly works

mgrunbauer (approved):

lukas-phaf (approved):

PR #417, opened 2024-11-18
Interval can now be used to configure the advertised time resolution for liveupdate

maartenplieger:
See doc/tutorials/Configure_solar_terminator.md for details

### DIFF
--- a/adagucserverEC/LayerTypeLiveUpdate/LayerTypeLiveUpdate.cpp
+++ b/adagucserverEC/LayerTypeLiveUpdate/LayerTypeLiveUpdate.cpp
@@ -105,16 +105,24 @@ int layerTypeLiveUpdateConfigureWMSLayerForGetCapabilities(MetadataLayer *metada
     metadataLayer->layerMetadata.title.copy(metadataLayer->dataSource->cfgLayer->Name[0]->value.c_str());
   }
   CTime timeInstance;
+
+  CT::string timeResolution = "PT10M";
+
+  for (auto dim : metadataLayer->layer->Dimension) {
+    if (dim->value.equals("time") && !dim->attr.interval.empty()) {
+      timeResolution = dim->attr.interval;
+    }
+  }
+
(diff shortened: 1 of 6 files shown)


adagucserverEC/LayerTypeLiveUpdate/LayerTypeLiveUpdate.cpp
```diff
@@ -105,16 +105,24 @@ int layerTypeLiveUpdateConfigureWMSLayerForGetCapabilities(MetadataLayer *metada
     metadataLayer->layerMetadata.title.copy(metadataLayer->dataSource->cfgLayer->Name[0]->value.c_str());
   }
   CTime timeInstance;
+
+  CT::string timeResolution = "PT10M";
```
belentorrente, line 109: Maybe mention somewhere in the documentation that this is the default time resolution

data/config/datasets/adaguc.tests.solarterminator.xml
```diff
@@ -22,6 +22,7 @@
     <DataPostProc algorithm="solarterminator"/>
     <Variable>solarterminator</Variable>
     <Styles>soltstyle</Styles>
+    <Dimension interval="PT1M">time</Dimension>
```
belentorrente: I would put 10M here (1M looks like overkill for something like the solar terminator)

belentorrente (dismissed):

adagucserverEC/LayerTypeLiveUpdate/LayerTypeLiveUpdate.cpp
maartenplieger: Wrote a comment in doc/configuration/Layer.md

data/config/datasets/adaguc.tests.solarterminator.xml
maartenplieger: Updated!

belentorrente (approved):

Issue #316, opened 2023-11-10
WMS problem with OpenDAP (and NetCDF files) with uneven latitude and/or longitude


MarcoAlbaETT:
Hi,
I've some NetCDF files with uneven latitude and/or longitude.
When I'll try to display these file via WMS , this is the result:
![image](https://github.com/KNMI/adaguc-server/assets/30118349/76a000bd-04fd-4954-9e1b-a1fa00ec2659)

I've tried different configuration, all with the same results:

- Using OpenDAP service from ERDDAP: https://prod-erddap.emodnet-physics.eu/erddap/griddap/MULTIOBS_GLO_PHY_SSS_L4_MY_015_015

- Using one of the NetCDF files used to build the dataset (downloadable here: https://prod-erddap.emodnet-physics.eu/erddap/files/MULTIOBS_GLO_PHY_SSS_L4_MY_015_015/cmems_obs-mob_glo_phy-sss_my_multi-oi_P1W/) placing the file in the "data" folder and also in the "autowms" folder.

The OpenDAP service used in ncWMS or Panoply is displaying the data in the right way.
Using the file in Panoply is also working as expected

There is a way to display these type of data in the right way?

Thanks

Marco

maartenplieger:
Hi Marco,

We have developed support for irregular grids, see https://github.com/KNMI/adaguc-server/releases

This might help with what you want! Can you try?

maartenplieger:
This is related to https://github.com/KNMI/adaguc-server/issues/174

MarcoAlbaETT:
Hi!
I've tried two different configuration


**1)** 
I've used this configiuration:
```
<?xml version="1.0" encoding="UTF-8" ?>
<Configuration>	
<Legend name="rainbow" type="colorRange">
<palette index="0" red="0" green="0" blue="255"/>
<palette index="80" red="0" green="255" blue="255"/>
<palette index="120" red="0" green="255" blue="0"/>
<palette index="160" red="255" green="255" blue="0"/>
<palette index="240" red="255" green="0" blue="0"/>
</Legend>
<Style name="rainbow">
<Legend fixedclasses="true" tickinterval="500"
tickround="1">rainbow</Legend>
<Min>0</Min>
<Max>30</Max>
</Style>	
  <Layer type="database">
    <FilePath>https://prod-erddap.emodnet-physics.eu/erddap/griddap/INSITU_GLO_PHY_TS_OA_MY_013_052</FilePath>  
[...]

PR #422, opened 2024-12-23
316 fix uneven lat lon grids

maartenplieger:
## Fixed issue with irrular grids for SSS SMOS L4 OI - LOPS-v2021 and Global Ocean - Coriolis Observation Re-Analysis CORA5.2. Added test 

Fixed by setting datatype to CDF_FLOAT for reading irregular grids. Some grids are in CDF_DOUBLE and are not working with current code.

Fixes Issue #316


NEWS.md
```diff
@@ -1,3 +1,6 @@
+**Version 2.29.2 2024-12-23**
+- Fixed issue with irrular grids for SSS SMOS L4 OI - LOPS-v2021 and Global Ocean - Coriolis Observation Re-Analysis CORA5.2. Added test (Issue #316)
```
belentorrente: typo: irr**eg**ular

belentorrente:
So if the grid of the dataset is in double precision, from a technical point of view the precision will be "downgraded" to float precision, right? Is this something that needs to be specified somewhere in the documentation for users?

NEWS.md
maartenplieger: Fixed

maartenplieger:
> So if the grid of the dataset is in double precision, from a technical point of view the precision will be "downgraded" to float precision, right? Is this something that needs to be specified somewhere in the documentation for users?

You are right that the data is downgraded.  In the visualisation it is impossible to see and in the Web Coverage Service you will see that the returned data is of type float. Furthermore, I don't think gridded data in CDF_DOUBLE makes sense...  I don't think we need to mention.

belentorrente (approved):

PR #428, opened 2025-02-17
Enabled test for AAIGRID WCS GetCoverage request in original projection

maartenplieger:
- Re-enabled previously disabled test.
- Should behave the same on different platforms and different GDAL versions


adagucserverEC/CGDALDataWriter.cpp
```diff
@@ -602,13 +537,10 @@ int CGDALDataWriter::end() {
...
   }
-
-  // CDBDebug("TEST %s?",driverName.c_str());
-
   if (driverName.equalsIgnoreCase("AAIGRID")) {
     CDBDebug("Setting FORCE_CELLSIZE to TRUE for AAIGRID");
-    papszOptions = CSLSetNameValue(papszOptions, "FORCE_CELLSIZE", "TRUE");
-    // papszOptions = CSLSetNameValue( papszOptions, "GDAL_VALIDATE_CREATION_OPTIONS", "FALSE" );
+    // papszOptions = CSLSetNameValue(papszOptions, "FORCE_CELLSIZE", "TRUE");
+    papszOptions = CSLSetNameValue(papszOptions, "GDAL_VALIDATE_CREATION_OPTIONS", "FALSE");
```
mgrunbauer: The `CDBDebug` still says `FORCE_CELLSIZE`, but you've commented this out. Could you change the log line?

Also could you add a short comment explaining why you need to set the `GDAL_VALIDATE_CREATION_OPTIONS=FALSE` option?

tests/AdagucTests/TestWCS.py
```diff
@@ -208,19 +209,45 @@ def test_WCSGetCoverageGeoTiff_testdatanc(self):
...
+        round(float(re.split(r'\s+', lines[6])[1]),3)]
+    
+    self.assertEqual([  'ncols', '29',
+                        'nrows', '31', 
+                        'xllcorner', -1500000.01, 
+                        'yllcorner', -999999.97, 
+                        'dx',        103448.28,
+                        'dy',        96774.19,
+                        'NODATA_value', -0.01],
+                     aaigrid_header_items)
```
mgrunbauer: This is a nicer way of parsing this data:

```python
lines = str(data.getvalue().decode('UTF-8')).splitlines()
parsed = [re.split(r'\s+', line) for line in lines]
aaigrid_header_items = [(p[0], round(float(p[1]), 3)) for p in parsed]

self.assertEqual([('ncols', 29.0), ('nrows', 31.0), ('xllcorner', -1500000.013), ('yllcorner', -4206896.551), ('cellsize', 103448.277), ('NODATA_value', -0.01)], aaigrid_header_items)
```

Note that this does round all numbers with 3 digits behind the dot. So `-1500000.01` will become `-1500000.013`. Also we parse the ints, so `'29'` becomes `29.0`.

mgrunbauer:
The test `test_WCSGetCoverageAAIGRID_NATIVE_testdatanc` works on macOS (outside docker) 👍

mgrunbauer (approved):

Issue #400, opened 2024-09-02
Docker build should run psql tests, requires psql service in github actions


mgrunbauer:
Adaguc supports both PostgreSQL and SQLite as database back-ends, however we want to deprecate SQLite and only support PostgreSQL.

The PR https://github.com/KNMI/adaguc-server/pull/361 adds the file `runtests_psql.sh` which can already run the tests against PostgreSQL during the docker build. This works when building adaguc on a local system. Note that this requires a running PostgreSQL server reachable on port 5432.

We should try using a PostgreSQL service in GitHub Actions. This would allow us to run all tests against PostgreSQL. If this works, we can remove support for SQLite.

PR #433, opened 2025-03-18
Exceeding maxquerylimit results in error 

### LINKED ISSUE #409: Move limit in getFilesAndIndicesForDimensions to time dimension
In https://github.com/KNMI/adaguc-server/pull/361, a PostgreSQL query in the function `getFilesAndIndicesForDimensions` was rewritten to this format:
 
```
select distinct t1.path,member,dimmember,height,dimheight,time,dimtime from
    t20240301t095153031_xkq9pp0nv3kqmonttvhf t1
inner join t20240301t095153038_hqug9msl1wgxpew4yozf t2
    ON t1.path = t2.path
inner join t20240301t095153043_oe2pyeutb8pdrvb7gu7r t3
    ON t1.path = t3.path
where
    member = 'member4' and
    t1.adaguctilinglevel != -1 and
    abs(5000 - height) = (select min(abs(5000 - height)) from t20240301t095153038_hqug9msl1wgxpew4yozf) and
    time = '2017-01-01T00:15:00Z'
order by member,height,time
limit 512
```

The original goal of having a `limit` at all was to prevent Adaguc from returning data for a very large time period. However the `limit 512` now limits all the results from the query.

If we now perform an EDR `cube` call on a custom dimension (Ensemble data with `members=*` or statistical data with `percentile=*`), EDR will respond with a limited set.

The result of the query is `time steps * custom dimension steps`:
- With 144 timesteps, you can only query 3 percentiles, because `144 * 3 = 432`, which fits inside 512.
- Querying 4 percentiles results in reduced timesteps, because `144 * 4 = 576`, which is more than 512.

We should rewrite the query, for example by using a subquery on the time dimension, with the limit, in the inner join. Something like the following (untested):

```
select distinct t1.path,member,dimmember,height,dimheight,time,dimtime from
    t20240301t095153031_xkq9pp0nv3kqmonttvhf t1
inner join (SELECT * FROM t20240301t095153038_hqug9msl1wgxpew4yozf t2 WHERE ... LIMIT 512)
    ON t1.path = t2.path
inner join t20240301t095153043_oe2pyeutb8pdrvb7gu7r t3
    ON t1.path = t3.path
where
    member = 'member4' and
    t1.adaguctilinglevel != -1 and
    abs(5000 - height) = (select min(abs(5000 - height)) from t20240301t095153038_hqug9msl1wgxpew4yozf) and
    time = '2017-01-01T00:15:00Z'
order by member,height,time
```

Testing and verifying this functionality is easiest after custom dimension support is added to EDR cube.

mgrunbauer:
When exceeding the maxquerylimit, by for example querying 60 members * 50 timesteps, Adaguc would normally return _some_ data, but not the complete dataset. This was confusing.

We decided to keep the `LIMIT=` in the same place to prevent the possibility of accidentally overloading Adaguc.

This MR checks if the limit gets exceeded, and throws a HTTP 422. The limit can still be raised by setting `maxquerylimit=<some big number>` in the `FilePath` element.
To be able to query _all_ the data, you should set the limit to the product of all the dimensions you want to query (e.g. `50 timesteps * 60 members = 3000`).

This closes https://github.com/KNMI/adaguc-server/issues/409.


tests/AdagucTests/TestWMSTimeSeries.py
```diff
@@ -366,4 +366,45 @@ def test_wmsgetfeatureinfo_json_timeheight_profile_for_sorted_model_levels(self)
...
             AdagucTestTools().readfromfile(self.expectedoutputsspath + filename),
-        )
\ No newline at end of file
+        )
+
+    def test_WMSGetFeatureInfo_exceed_maxquerylimit(self):
+        AdagucTestTools().cleanTempDir()
+        if os.getenv("ADAGUC_DB", "").endswith(".db"):
+            print("SKIP: Only PSQL")
+            return
```
mgrunbauer, line 375: The test works when running in PostgreSQL, however we run the tests in the docker container with sqlite and we're using a different query there which then fails this test.

[This issue](https://github.com/KNMI/adaguc-server/issues/400) mentions running the tests in Docker against PostgreSQL (and possibly deprecating/removing sqlite support?). Maybe we can then remove these `if psql: return` checks.

data/datasets/members.nc
mgrunbauer, line 1: This is a test file with the dimensions:
```
dimensions:
	lat = 3 ;
	lon = 3 ;
	time = 60 ;
	member = 50 ;
```

maartenplieger (approved):

PR #435, opened 2025-03-19
Add support for odim hdf5 volume scans

mats-knmi:
In this pull request I extended knmi hdf5 volume data support added in https://github.com/KNMI/adaguc-server/pull/434 to also allow for ODIM hdf5 files to be visualized. Since both visualizations share a lot of the same code I added a utils file with methods to retrieve the data and metadata from the hdf5 file. These methods are set up so that they detect what file type is supplied and then retrieve the correct information automatically. This allows the visualization code to be unaware of the file type of the data it is visualizing, so that that code can focus on the visualization only.


adagucserverEC/CConvertH5VolScanUtils.cpp
```diff
@@ -0,0 +1,282 @@
+/******************************************************************************
+ *
+ * Project:  ADAGUC Server
+ * Purpose:  Utils for conversion HDF5 volume scan data to CDM
+ * Author:   Ernst de Vreede, ernst.de.vreede "at" knmi.nl
+ * Date:     2022-08-15
```
belentorrente: Just a minor thing but, is this really a file from 2022? I am assuming you copied the info and copyright statement from another file and forgot to change this and also the author.

mats-knmi: Yes you are right. I actually made multiple copies of this file during this work and at first I was updating this information, but in the final copy I forgot... I pushed a commit that fixes this

maartenplieger (approved):

PR #437, opened 2025-03-19
Simplify legend when there are too many discrete classes

belentorrente:
This MR deals with the situation where the legend becomes impossible to see due to the presence of too many classes. 

The assumption in this special case is that we are working with temperatures. The simplified version of the legend will show the legends of multiples of 5 (it checks every class for a multiple of 5, and if it's in the range, a label with the min of said interval will be added). 

For now this mainly affects the ha43 DINI temperature layers, so this is the data used in the test.

Also added a test for the `temperature_wow` style, because there was no test case for that part of the code.


adagucserverEC/CDrawImage.cpp
```diff
@@ -1426,11 +1426,15 @@ int CDrawImage::createGDPalette(CServerConfig::XMLE_Legend *legend) {
   return 1;
 }
 
-void CDrawImage::rectangle(int x1, int y1, int x2, int y2, CColor innercolor, CColor outercolor) {
+void CDrawImage::rectangle(int x1, int y1, int x2, int y2, CColor innercolor, CColor outercolor, bool noVerticalSpace) {
```
maartenplieger: I think it would be better to do this calculation in where you call rectangle.

It should be possible to achieve what you want without touching the draw functions.

tests/expectedoutputs/TestWMS/test_WMSGetLegendGraphic_LongTempLegend.png
maartenplieger, line 1: Looks nice!

adagucserverEC/CDrawImage.cpp
belentorrente: Maybe I can play with the size of the block to be one pixel longer in height

belentorrente: Yes, it was easier to fix than I thought. You can have a look again.

adagucserverEC/CLegendRenderers/CCreateLegendRenderDiscreteLegend.cpp
```diff
@@ -129,40 +139,100 @@ int CCreateLegend::renderDiscreteLegend(CDataSource *dataSource, CDrawImage *leg
...
+      char szTemp[1024];
+      size_t numShadeIntervals = styleConfiguration->shadeIntervals->size();
+
+      // Calculate which part of the legend to draw (only between min and max)
+      int minInterval = 0;
+      int maxInterval = numShadeIntervals;
+      for (size_t j = 0; j < numShadeIntervals; j++) {
+        CServerConfig::XMLE_ShadeInterval *s = (*styleConfiguration->shadeIntervals)[j];
+        if (!s->attr.min.empty() && !s->attr.max.empty()) {
+          float intervalMinf = parseFloat(s->attr.min.c_str());
```
maartenplieger: Would be nice to add a property inside the element RenderSettings to indicate wether the legend should show its full color range or only the clipped colors according to the data. See https://github.com/KNMI/adaguc-server/blob/master/adagucserverEC/CServerConfig_CPPXSD.h#L809C103-L809C144

adagucserverEC/CLegendRenderers/CCreateLegendRenderDiscreteLegend.cpp
```diff
@@ -129,40 +139,100 @@ int CCreateLegend::renderDiscreteLegend(CDataSource *dataSource, CDrawImage *leg
...
+    // If this type of legend has too many classes (15+) with a relatively small height (<=500 px), we
+    // will treat it differently to create a summarised version, with the following characteristics:
+    // - The class border is removed for extra visibility
+    // - Labels are simplified and will only show the min of the interval the class represents
+    // - Only one every 5 labels will be printed (this is just a rule of thumb) plus the top label
+    // - Only classes the min and the max data value will be added
+    const size_t MAX_DISCRETE_CLASSES = 15;
+
+    // Case where the legend has too many classes to plot
+    if (styleConfiguration->shadeIntervals->size() > MAX_DISCRETE_CLASSES && legendImage->Geo->dHeight <= 500) {
```
maartenplieger: Maybe check if this can be set by a flag via RenderSettings or check if it can be deteriminded by blockheight.

maartenplieger (approved):

PR #441, opened 2025-03-20
add launch.json for vscode debugging

mats-knmi:


.vscode/launch.json
```diff
@@ -0,0 +1,84 @@
...
+                    "name": "ADAGUC_ENABLELOGBUFFER",
+                    "value": "FALSE"
+                },
+                {
+                    "name": "QUERY_STRING",
+                    "value": "service=WMS&request=GetCapabilities"
+                }
+            ],
+            "externalConsole": false,
+            "MIMode": "gdb",
```
mgrunbauer: `gdb` is not available for me under macos, I was able to use `lldb` though. Does the following change work for you too under linux?

```suggestion
            "linux": {
                "MIMode": "gdb",
            },
            "osx": {
                "MIMode": "lldb"
            },
```

mgrunbauer (dismissed):
Nice, I am impressed by this 👍 

I've tested the python debugger by setting a breakpoint in the EDR code and doing a `curl http://localhost:8080/edr/collections`. And I've tested the C++ debugger by putting a breakpoint in the png code and setting the QUERY_STRING to a `GetMap` request. Works great!

I did have to set `lldb` under macos, could you check if my suggested change still works for you under linux?

.vscode/launch.json
belentorrente: @mgrunbauer I tried out your change and it also works!

mats-knmi: Yes is still works for me as well!

mats-knmi:
@mgrunbauer or @belentorrente, I bumped the version and applied the proposed change, so this is ready to be merged. If one of you can approve it again and click merge then this PR can be done.

belentorrente (approved):

PR #448, opened 2025-03-27
Run tests against PostgreSQL by default

### LINKED ISSUE #400: Docker build should run psql tests, requires psql service in github actions

mgrunbauer:
When building the Adaguc container, we also run the testsuites which need a database to work. The default is SQLite, however there are lots of tests which require PostgreSQL, these tests currently get skipped.

This MR:
- Makes the tests run against PostgreSQL by default.
- Adds an extra `docker-compose-test.yml` which includes a postgres database, to not clash with your local setup.
- The test database runs against port `54321` by default.
- Adds an "is psql running?" check in `starttests_psql.sh` and shows how to setup a local database if needed.

There are three different situations in which the tests can connect to PSQL:
- The build in github actions uses an extra `postgres` service in the CI/CD. The tests will automatically connect to psql on `localhost`, which works in github actions (with `network=host`).
- Doing `docker build ...` on your local machine, the tests will connect to psql on `host.docker.internal`. This should get tested on both linux and macOS (@lukas-phaf). It currently works for me :)
- Doing `bash runtests_psql.sh` on your local machine, the tests will connect to psql on `localhost`.

This could be a first step to deprecate SQLite in Adaguc.
This MR will close https://github.com/KNMI/adaguc-server/issues/400.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -22,6 +22,18 @@ jobs:
             runner: ubuntu-24.04
           - platform: linux/arm64
             runner: ubuntu-24.04-arm
+
+    services:
+      postgres:
+        image: postgres:13
+        env:
+          POSTGRES_USER: adaguc
+          POSTGRES_PASSWORD: adaguc
+          POSTGRES_DB: adaguc
+        ports:
+        - 5432:5432
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+
     steps:
       - name: Prepare
         run: |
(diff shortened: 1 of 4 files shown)


ernstdevreede (approved):
Works for me

PR #456, opened 2025-04-16
Allow dbztorr postproc to handle both float and double data

mats-knmi:


adagucserverEC/CDataPostProcessors/CDataPostProcessor_CDPDBZtoRR.cpp
```diff
@@ -27,38 +25,35 @@ int CDPDBZtoRR::execute(CServerConfig::XMLE_DataPostProc *proc, CDataSource *dat
...
-          data[j] = getRR(data[j]);
+          data[j] = pow((pow(10., data[j] * 0.1) / 200.), 0.625);
         }
       }
     }
   }
   return 0;
 }
 
+int CDPDBZtoRR::execute(CServerConfig::XMLE_DataPostProc *proc, CDataSource *dataSource, int mode, double *data, size_t numItems) { //
```
ernstdevreede, line 40: I do not see where this methos is used.
Is it necessary?

ernstdevreede (commented):
One remark

adagucserverEC/CDataPostProcessors/CDataPostProcessor_CDPDBZtoRR.cpp
mats-knmi: It is not used, but since this class is a subclass of CDPPInterface, which has this method, I had to implement it...

ernstdevreede: Ah, that's why the method is there.
I was just wondering why there was this method, that only handles double data types.

ernstdevreede (approved):

PR #457, opened 2025-04-23
Don't reuse generated table name

### LINKED ISSUE #408: BUG: Configuring a databasetable in adaguc-server does not seem to work anymore
The following xml file does not seem to work anymore. See attached HDF5 files in second message as well.

When you remove the DataBaseTable, the layer starts to work again.

If you configure a database table, the layer stops working.

```xml

<?xml version="1.0" encoding="UTF-8"?>
<Configuration>
  <Include location="/data/adaguc-datasets/includes/StandardStyles_RADAR.xml" />

  <Environment name="ADAGUCENV_RETENTIONPERIOD" default="P7D" />
  <Environment name="ADAGUCENV_ENABLECLEANUP" default="true" />
  <Settings enablecleanupsystem="{ADAGUCENV_ENABLECLEANUP}" cache_age_volatileresources="60" cache_age_cacheableresources="36000"/>


  <Layer type="database">
    <Name force="true">3d_composite_10</Name>
    <Title>3d composite dBZ&gt;-10</Title>
    <Group value="3D_RADAR_COMPOSITE"/>
    <FilePath filter="^3D_RADAR_COMPOSITE.*\.nc$">/data/adaguc-data/RADAR/EWC/3d-radar-composite/</FilePath>
    <Variable>reflectivity</Variable>
    <Dimension name="time" interval="PT5M" units="ISO8601">time</Dimension>
    <Dimension name="z" units="km">z</Dimension>
    <Styles>dBZ_more_than_-10_3D</Styles>
    <DataBaseTable>RADAR_COMPOSITE_3D_2</DataBaseTable>
  </Layer>
  <Layer type="database">
    <Name force="true">3d_composite_0</Name>
    <Title>3d composite dBZ&gt;0</Title>
    <Group value="3D_RADAR_COMPOSITE"/>
    <FilePath filter="^3D_RADAR_COMPOSITE.*\.nc$">/data/adaguc-data/RADAR/EWC/3d-radar-composite/</FilePath>
    <Variable>reflectivity</Variable>
    <Dimension name="time" interval="PT5M" units="ISO8601">time</Dimension>
    <Dimension name="z" units="km">z</Dimension>
    <Styles>dBZ_more_than_0_3D</Styles>
    <DataBaseTable>RADAR_COMPOSITE_3D_2</DataBaseTable>
  </Layer>
  

  <!-- End of configuration /-->
</Configuration>

```

mgrunbauer:
When generating the table name, we were appending dimension names to the preconfigured table name. We then reused the changed tablename.

For example, with a `<DataBaseTable>MYTABLE</DataBaseTable>` and the dimensions `mydim1`, `mydim2` and `mydim3`, this resulted in adaguc trying to query the following tables:

- mytable_mydim1
- mytable_mydim1_mydim2
- mytable_mydim1_mydim2_mydim3

... which is wrong and should have been:

- mytable_mydim1
- mytable_mydim2
- mytable_mydim3

This also explains why it did work for certain datasets, these only had a single dimension configured.

I've added a testcase in `TestWMS.py`. It isn't exactly tied to WMS specifically but it was the easiest way to prove it by just trying to generate an image for a specific dataset.

This fixes https://github.com/KNMI/adaguc-server/issues/408


adagucserverEC/CDBAdapterPostgreSQL.cpp
```diff
@@ -1144,7 +1149,7 @@ CDBStore::Store *CDBAdapterPostgreSQL::getLayerMetadataStore(const char *dataset
 
   if (this->layerMetaDataStore == nullptr) {
 #ifdef CDBAdapterPostgreSQL_DEBUG
-    CDBDebug("Need to query layer metadata for %s/%s/%s", datasetName, layerName, metadataKey);
```
mgrunbauer, line 1147: When setting `CDBAdapterPostgreSQL_DEBUG`, compiling failed because we don't have `layerName` and `metadataKey` present anymore (removed during metadata refactors).

ernstdevreede (dismissed):

lukas-phaf:
Just to be clear, this only affects the table name that is used for reading, correct?
So this doesn't change where the table data is actually stored, potentially breaking existing tables?

adagucserverEC/CDBAdapterPostgreSQL.cpp
```diff
@@ -666,9 +672,8 @@ std::map<CT::string, DimInfo> CDBAdapterPostgreSQL::getTableNamesForPathFilterAn
 
   // If config has a hardcoded db table name for layer, use it
   if (dataSource->cfgLayer->DataBaseTable.size() == 1) {
-    CT::string tableName = dataSource->cfgLayer->DataBaseTable[0]->value.c_str();
-
     for (auto &dim : dimensions) {
+      CT::string tableName = dataSource->cfgLayer->DataBaseTable[0]->value.c_str();
       dataSource->srvParams->makeCorrectTableName(&tableName, &dim);
```
maartenplieger: Can you rewrite 
```

void CServerParams::makeCorrectTableName(CT::string *tableName, CT::string *dimName) {
  tableName->concat("_");
  tableName->concat(dimName);
  tableName->replaceSelf("-", "_m_");
  tableName->replaceSelf("+", "_p_");
  tableName->replaceSelf(".", "_");
  tableName->toLowerCaseSelf();
}

```

to 

```
CT::string CServerParams::makeCorrectTableName(CT::string tableName, CT::string dimName) {
  CT::string correctedTableName;
  correctedTableName.print("%s_%s", tableName.c_str(), dimName.c_str();
  correctedTableName->replaceSelf("-", "_m_");
  correctedTableName->replaceSelf("+", "_p_");
  correctedTableName->replaceSelf(".", "_");
  correctedTableName->toLowerCaseSelf();
  return correctedTableName
}

```

This will make it more functional.

Then you should also use:

```
 for (auto &dim : dimensions) {
      CT::string tableName = dataSource->cfgLayer->DataBaseTable[0]->value.c_str();
      CT::string correctedTableName = dataSource->srvParams->makeCorrectTableName(tableName, dim);
      mapping[dim] = {correctedTableName, ""};
    }
```

maartenplieger: Tested it out, and it works as expected. Nice find! Also there is no changed functionality for GeoWeb/KDP. 

Actually, layers with multiple dimensions and a databasetable now start to work.

:clap:

mgrunbauer:
> Just to be clear, this only affects the table name that is used for reading, correct? So this doesn't change where the table data is actually stored, potentially breaking existing tables?

Yes, that is correct.

On master, scanning a new dataset with multiple dimensions and the following dataset configuration:
```xml
<Configuration>
  <Layer type="database">
    <FilePath>/data/adaguc-data/multi_dim.nc</FilePath>
    <Variable>testdata</Variable>
    <Name>my_testdata</Name>
    <Styles>auto</Styles>
    <Dimension name="time" units="ISO8601" interval="PT1H" default="min">time</Dimension>
    <Dimension name="height" units="m" type="vertical">height</Dimension>
    <Dimension name="forecast_reference_time" units="ISO8601">reference_time</Dimension>
    <DataBaseTable>MY_TABLE</DataBaseTable>
  </Layer>
</Configuration>
```
results in the following tables getting created:
```psql
 adaguc=# \d
                     List of relations
 Schema |               Name               | Type  | Owner
--------+----------------------------------+-------+--------
 public | layermetadata                    | table | adaguc
 public | my_table_forecast_reference_time | table | adaguc
 public | my_table_height                  | table | adaguc
 public | my_table_time                    | table | adaguc
 public | pathfiltertablelookup_v2_0_23    | table | adaguc
(5 rows)
```

Querying (e.g. by doing a `GetMap`) now fails on master because adaguc tries to query the table `my_table_forecast_reference_time_height_time`.

With this fix and _without_ dropping any tables or re-scanning, adaguc uses the correct table names (shown in the output above).

Dropping and re-scanning results in the same (already correct) table names getting created.

maartenplieger (approved):

PR #470, opened 2025-06-18
Fix for memleak when using variable rename (and two other small ones)

lukas-phaf:


adagucserverEC/CDFObjectStore.cpp
```diff
@@ -45,6 +46,25 @@ const char *CDFObjectStore::className = "CDFObjectStore";
 extern CDFObjectStore cdfObjectStore;
 CDFObjectStore cdfObjectStore;
 bool EXTRACT_HDF_NC_VERBOSE = false;
+
+
+std::string gen_random_string( size_t length )
```
maartenplieger: There is a 
```
  static const CT::string randomString(const int len);
```

adagucserverEC/CDFObjectStore.cpp
```diff
@@ -309,9 +329,8 @@ CDFObject *CDFObjectStore::getCDFObject(CDataSource *dataSource, CServerParams *
         if (!cfgVar->attr.orgname.empty()) {
           CDF::Variable *var = cdfObject->getVariable(cfgVar->attr.orgname);
           var->name.copy(cfgVar->value);
-          // This cdfobject should not be cached, as the variable
-          // can no longer be found by other layers using the same cdfobject.
-          cached = false;
+          // HACK: We want it in the cache (so the store is responsible for cleaning it up), but we don't want it reused.
+          uniqueIDForFile = uniqueIDForFile + "_" + gen_random_string(10).c_str();
```
maartenplieger: You could use `CServerParams::randomString(10).c_str()`

adagucserverEC/CDFObjectStore.cpp
```diff
@@ -45,6 +46,25 @@ const char *CDFObjectStore::className = "CDFObjectStore";
...
+
+std::string gen_random_string( size_t length )
+{
+  auto randchar = []() -> char
+  {
+    const char charset[] =
+        "0123456789"
+        "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+        "abcdefghijklmnopqrstuvwxyz";
+    const size_t max_index = (sizeof(charset) - 1);
```
maartenplieger: Length should be one  less (-2)

maartenplieger (approved):

PR #471, opened 2025-06-18
Point text outline

loescornelis:
Allow an outline for the rendering of point data as text (pointstyle="point" where discradius = 0).


adagucserverEC/CCairoPlotter.cpp
```diff
@@ -682,6 +682,13 @@ void CCairoPlotter::drawStrokedText(int x, int y, double angle, const char *text
   cairo_new_path(cr);
   cairo_set_dash(cr, 0, 0, 0);
 
+  if (centerText) {
+    cairo_text_extents_t te;
+    cairo_text_extents(cr, text, &te);
+    x = x - (te.x_bearing + te.width / 2);
+    y = y = y - (te.y_bearing + te.height / 2);
```
maartenplieger: Too many y=y=y?

loescornelis: Fixed!

ernstdevreede (approved):

PR #477, opened 2025-06-26
Number alignment and other legend improvements

belentorrente:
This MR includes

- Alignment of legend numbers around the decimal point (or right aligned otherwise).
- Now legends can represent any UTF-8 character (including the _en dash_ for intervals).
- Font used in Legends is now Open Sans (for it's relatively modern and light look). 
- Roboto set as contour font instead of Arial


adagucserverEC/CCairoPlotter.cpp
```diff
@@ -292,6 +292,25 @@ int CCairoPlotter::initializeFreeType() {
   return 0;
 }
 
+// Aux function to support unicode characters (such as en dash)
+const char *decode_utf8_char(const char *p, uint32_t *out_char) {
```
maartenplieger: Is this really needed? I think it could potentially cause buffer overflows.

data/config/datasets/adaguc.tests.nearestshadeinterval.xml
```diff
@@ -2,7 +2,7 @@
 <Configuration>
   <!-- Custom styles -->
   <Style name="shadedstylefast">
-    <ShadeInterval min="-80" max="-7" fillcolor="#08387b" bgcolor="#FF000080" />
+    <ShadeInterval min="-8" max="-7" fillcolor="#08387b" bgcolor="#FF000080" />
```
maartenplieger: The min -80 is on purpose.  This is ensures that the lowest class gets coloured completely. You can you a label instead icm the min on -80.

This will also fix your test in adaguc.tests.nearestshadeinterval.xml

data/config/datasets/adaguc.tests.nearestshadeinterval.xml
```diff
@@ -61,7 +61,7 @@
 
 
   <Style name="shadedstyleprecise">
-    <ShadeInterval min="-80" max="-7" fillcolor="#08387b" bgcolor="#FF000080" />
+    <ShadeInterval min="-8" max="-7" fillcolor="#08387b" bgcolor="#FF000080" />
```
maartenplieger: Same here

tests/expectedoutputs/TestWMS/test_WMSGetMap_dashed_contour_lines.png
maartenplieger, line 1: Is it possible to add a test with the previous font as well?

adagucserverEC/CCairoPlotter.cpp
belentorrente: I can look into other (safer) ways of doing this. I would like to keep the UTF-8 charset, it is a nice addition. Makes it more readable to write a long dash instead of a short one. (so that it does not get confused for negative numbers etc). I will fix the other issues first

data/config/datasets/adaguc.tests.nearestshadeinterval.xml
belentorrente: Very nice! I couldn't have guessed this was related (because the images with the pinkness do not even have a legend, so I didn't "connect the dots")

tests/expectedoutputs/TestWMS/test_WMSGetMap_dashed_contour_lines.png
belentorrente: Like, configuring it to use arial instead?

data/config/datasets/adaguc.tests.nearestshadeinterval.xml
belentorrente: I fixed this by added extra checks so it doesn't go out of bounds

tests/expectedoutputs/TestWMS/test_WMSGetMap_dashed_contour_lines.png
belentorrente: Arial tests are working

maartenplieger (approved):

PR #478, opened 2025-06-26
Cache meta

ernstdevreede:
This PR adds caching of getmetadata calls (if redis URL is configured).
The metadate is written to the REDIS cache each time after the metadata has been updated and the cached info is used by the code for quickly retrieving the metadata (for example for EDR /collections calls).


ernstdevreede:
Caching is considered harmful.
Closed

PR #483, opened 2025-06-30
More edr locations

ernstdevreede:
This MR adds "more EDR locations" to EDR services of ADAGUC.
A standard list of stations is provided in the source code (in the file $ADAGUC_PATH/data/resources/locations/global_edr_locations.geojson, which could be overridden in real deployments).

The list of stations is filtered by the LatLonBox from the GetMetadata call for the collection, when /locations are requested. If more than 0 locations remain after filtering the /locations call is enabled and the remaining locations are usable for /locations/LOCATION_ID calls.


maartenplieger (approved):

python/python_fastapi_server/routers/edr_locations.py
```diff
@@ -9,37 +9,76 @@
...
-@router.get("/collections/{_coll}/locations")
-@router.get("/collections/{_coll}/instances/{instance}/locations")
-@router.get("/collections/{_coll}/locations/{id}")
-@router.get("/collections/{_coll}/instances/{instance}/locations/{location_id}")
+async def get_locations_for_collection(coll: str):
+    metadata = await get_metadata(coll)
+
+    locations_for_coll = []
+    try:
+        bbox = metadata[coll][next(iter(metadata[coll]))]["layer"].get("latlonbox")
```
lukas-phaf, line 56: Hard to read... but first layer is taken as leading?

Instead of getting the first key and stuffing that back in the dict, why not just get the first entry? Something like:
```
next(iter(metadata[coll].values()))["layer"].get("latlonbox")
```

python/python_fastapi_server/routers/edr_locations.py
```diff
@@ -9,37 +9,76 @@
...
+        for location in get_edr_locations():
+            coords = location["geometry"]["coordinates"]
+            if coords[1] >= bbox[1] and coords[1] <= bbox[3]:
```
lukas-phaf, line 59: Python allows you to write`a<=x<=b`

python/python_fastapi_server/routers/edr_locations.py
```diff
@@ -9,37 +9,76 @@
...
+                if coords[0] >= bbox[0] and coords[0] <= bbox[2]:
+                    locations_for_coll.append(location)
+        return locations_for_coll
+    except KeyError:
+        return GLOBAL_LOCATIONS
```
lukas-phaf, line 64: We do want to log an error here...

python/python_fastapi_server/routers/edr_locations.py
```diff
@@ -9,37 +9,76 @@
...
 
 router = APIRouter()
 logger = logging.getLogger(__name__)
 logger.debug("Starting EDR")
 
+GLOBAL_LOCATIONS = []
+
+
+def get_edr_locations():
+    global GLOBAL_LOCATIONS
```
lukas-phaf, line 28: Why not use functools.cache or lru_cache instead?

PR #493, opened 2025-08-06
Refactored and made autowms async

mgrunbauer:
- The autowms calls where not async, this is now done through `await asyncio.to_thread(handler)`. I can do a call to `GetMap` _and_ `request=getfiles&path=/adaguc::data/dir-with-files` at the same time, and the `GetMap` call does not get blocked anymore.
- `handle_data_route` and `handle_autowmsdir_route` where nearly identical. I've merged these together.
- We did two `os.listdir` calls and some extra looping, I've rewritten this so it loops through `os.scandir` once and immediately builds the return data structure. Listing a path of 30k files is >2x as fast (on my local machine).
- Did some extra refactoring, added typehints, return the same `HTTPException` as we do in other endpoints.

### DIFF
--- a/python/python_fastapi_server/routers/autowms.py
+++ b/python/python_fastapi_server/routers/autowms.py
@@ -1,10 +1,13 @@
 """autoWmsRouter"""
+
+import asyncio
+from functools import partial
 import json
 import logging
 import os
 import urllib.parse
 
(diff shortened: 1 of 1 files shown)


maartenplieger (approved):
Test it and it works really well.

Issue #498, opened 2025-09-08
BUG: ADAGUC does not clean tables properly when used with files with multiple variables


maartenplieger:
Considering the following configuration, it seems that not all related tables get properly cleaned.

```xml
<?xml version="1.0"?>
<Configuration>

  <Include location="/data/adaguc-datasets/includes/StandardStyles_WIK.xml"/>
  <Environment name="ADAGUCENV_RETENTIONPERIOD" default="P1D"/>
  <Environment name="ADAGUCENV_ENABLECLEANUP" default="true"/>
  <Settings enable_edr="true" enablecleanupsystem="{ADAGUCENV_ENABLECLEANUP}" cleanupsystemlimit="1000" cache_age_volatileresources="60" cache_age_cacheableresources="36000"/>
  
  <WMS>
    <WMSFormat name="image/png" format="image/webp" />
  </WMS>
  
  <Layer type="database">
    <Name>air-pressure-ml</Name>
    <Variable>air-pressure-ml</Variable>
    <FilePath maxquerylimit="2048" retentionperiod="{ADAGUCENV_RETENTIONPERIOD}" retentiontype="datatime" filter="uwcw_ha43_dini_5p5km_soundings_geoweb_nc.*\.nc$">/data/adaguc-data/arcusuwcw_soundings/</FilePath>
    <Styles>pressure_cwk</Styles>
    <Dimension name="forecast_reference_time" units="ISO8601">reference_time</Dimension>
    <Dimension name="time" units="ISO8601" interval="PT1H" default="min" type="time">time</Dimension>
    <Dimension name="pml_at_ml" units="-" default="max" type="vertical">elevation</Dimension>
    <DataBaseTable>pmltable</DataBaseTable>
  </Layer>

  <Layer type="database">
    <Name>air-temperature-ml</Name>
    <Variable>air-temperature-ml</Variable>
    <FilePath maxquerylimit="2048" retentionperiod="{ADAGUCENV_RETENTIONPERIOD}" retentiontype="datatime" filter="uwcw_ha43_dini_5p5km_soundings_geoweb_nc.*\.nc$">/data/adaguc-data/arcusuwcw_soundings/</FilePath>
    <Styles>temperature,temperature_wow</Styles>
    <Dimension name="forecast_reference_time" units="ISO8601">reference_time</Dimension>
    <Dimension name="time" units="ISO8601" interval="PT1H" default="min" type="time">time</Dimension>
[...]

PR #499, opened 2025-09-08
Fix bug BUG ADAGUC does not clean tables properly

maartenplieger:
- Fix bug BUG ADAGUC does not clean tables properly when used with files with multiple variables [#498](https://github.com/KNMI/adaguc-server/issues/498)
- Also removed unneeded dbadapter


ernstdevreede (dismissed):
Works for me, I copmared old and new versions, cleanup now also clreas all database entries as fas as I can see.

ernstdevreede (approved):

PR #505, opened 2025-09-15
Edr loc

ernstdevreede:
This PR implements the EDR /locations/<ID> call by internally handling a /position call for the coordinates of the requested location.

I an unknown location ID is used (not configured in data/resources/locations/global_edr_locations.geosjon), the /locations/ID call returns a 404.

This PR also adds 2 includes of cstdint for compiling with newer GCC compilers.

This was previously implemented by redirecting to the /position call, but the new implementation saves a roundtrip for the HTTP redirect, which is noticable on cloud deployments


python/python_fastapi_server/routers/edr_locations.py
```diff
@@ -32,9 +39,10 @@ def get_edr_locations():
     all_locations = []
     try:
         with open(locations_file_path, "r", encoding="utf-8") as loc_f:
-            feature_collection = json.load(loc_f)
-            all_locations = feature_collection["features"]
-    except OSError:
+            # feature_collection = json.load(loc_f)
```
maartenplieger: Commented line can be removed?

python/python_fastapi_server/routers/edr_locations.py
```diff
@@ -62,39 +71,70 @@ async def get_locations_for_collection(coll: str):
...
+    allowed_params = ["datetime", "parameter-name", "z", "f"]
+    custom_dims = get_custom(
+        request.query_params,
+        allowed_params,
+    )
 
-    # Redirect to /position call with coordinates filled in
-    # in coords=POINT() argument if location id is known
-    req_url = str(request.url)
     for loc in location_list:
```
maartenplieger, line 124: Suggestion:
```
    found_location = next(filter(lambda x: x.id == location_id, location_list), None)
    if found_location:
        coords = f"POINT({found_location.geometry.coordinates[0]} {found_location.geometry.coordinates[1]})"
        return await handle_coll_inst_position(
            collection_name,
            custom_dims,
            coords,
            response,
            instance,
            datetime_par,
            parameter_name_par,
            z_par,
        )
```

Dockerfile
```diff
@@ -6,7 +6,7 @@ USER root
 LABEL maintainer="adaguc@knmi.nl"
 
 # Version should be same as in Definitions.h
-LABEL version="4.3.0"
+LABEL version="4.3.1"
```
maartenplieger, line 9: Can you update the NEWS.md also?

python/python_fastapi_server/routers/edr_locations.py
ernstdevreede: Done

Dockerfile
ernstdevreede: Done

python/python_fastapi_server/routers/edr_locations.py
ernstdevreede: Done

maartenplieger (approved):

PR #516, opened 2025-09-25
Use pytest and allow integrating in VS Code test explorer

### LINKED ISSUE #515: Allow running of individual adaguc functional tests in vscode
It would save time if we could run single tests easily from vscode.

Since we use pytest as the framework this should be possible.

mgrunbauer:
**Pytest**

This MR allows running the black-box tests and python (EDR) tests all through pytest.
Fixes https://github.com/KNMI/adaguc-server/issues/515.

Pytest uses a `conftest.py` which takes care of the testing setup:
- Setting the correct environment variables required for tests
- Setting up a testing database
- Setting `ulimit -c unlimited`

The following files have been removed because they are now unused:
```
./tests/starttests.sh
./tests/starttests_psql.sh
./tests/functional_test.py
./python/python_fastapi_server/test_server.sh
```

Only `./runtests_psql.sh` still exists. This is what the `docker build` uses to run the tests. It just calls `pytest` directly.

**VS Code**
- Added `.vscode/extensions.json`, VS Code will check if you have these extensions installed.
- After installing the python and CMake extensions, you can open the test explorer. The adaguc tests should appear there. Tests can be executed independently, and debugging the python code (and python tests) should also work. See screenshot for how it might look.
Screenshot 2025-10-08 at 16 51 46


tests/AdagucTests/TestADAGUCFeatureFunctions.py
```diff
@@ -12,20 +12,19 @@
...
     Run test for ADAGUC Feature functions
     """
-    testresultspath = "{ADAGUC_PATH}/tests/testresults/TestADAGUCFeatureFunctions/"
-    expectedoutputsspath = "{ADAGUC_PATH}/tests/expectedoutputs/TestADAGUCFeatureFunctions/"
-    env = {'ADAGUC_CONFIG': ADAGUC_PATH +
-           "/data/config/adaguc.autoresource.xml"}
-    testresultspath = testresultspath.replace("{ADAGUC_PATH}/", ADAGUC_PATH)
-    expectedoutputsspath = expectedoutputsspath.replace(
-        "{ADAGUC_PATH}/", ADAGUC_PATH)
+
```
maartenplieger: Please revert, the old is working

tests/starttests_psql.sh
```diff
@@ -1,5 +1,12 @@
 #!/bin/bash
 echo "Starting adaguc-server functional tests"
+
+# pytest \
```
maartenplieger: I think we should move towards this. And remove functionaltest.py.

tests/AdagucTests/TestGeoJSONFeatures.py
```diff
@@ -10,9 +10,12 @@
 import re
 from adaguc.AdagucTestTools import AdagucTestTools
 
-ADAGUC_PATH = os.environ['ADAGUC_PATH']
+import pytest
 
+ADAGUC_PATH = os.environ["ADAGUC_PATH"]
 
+
+@pytest.mark.skip("These tests are never called")
```
mgrunbauer, line 18: This file was not included in the `functional_test.py`. I've decided to skip these tests for now (they all fail).

.vscode/extensions.json
```diff
@@ -0,0 +1,8 @@
+{
+    "recommendations": [
+        "ms-python.python",
+        "ms-vscode.cmake-tools",
+        "ms-vscode.cpptools-extension-pack",
+        "ms-vscode.cpptools",
```
mgrunbauer, line 6: VS Code will ask you to install these extensions, and they are required to compile adaguc and run the python tests through test explorer. They are also mentioned in the `doc/developing/developingwithvscode.md`.

tests/AdagucTests/TestGeoJSONFeatures.py
maartenplieger: Yes that is fine

.vscode/extensions.json
maartenplieger: Useful!

maartenplieger (approved):

PR #528, opened 2025-10-27
Found and fixed memory issue in PNG reader

### LINKED ISSUE #529: ADAGUC crashes sometimes when reading PNG files
It seems that PNG reading can still cause Segmentation faults. 

Fix it this. E.g. by using the sanitzeadress option of the compiler.

maartenplieger:
- Use splitToStack instead of splitToArray (more robust, less null pointers, self cleaning memory and less code)
- Applied more functional programming instead of using static methods in classes. Also converted classes to structs.
- Refactor of BoundingBox to f8box
- Refactor of P to f8point
- Ran tiling with  `-ggdb -Wall -Wextra -g3 -fsanitize=address` to find all memory issues
- Found and fixed memory issue in PNG reader
- Fixed all compiler warnings on gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0

Closes https://github.com/KNMI/adaguc-server/issues/529

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+**Version 4.3.6 - 2025-10-27**
+
+- Making tiles does not cause memory violations
+
 **Version 4.3.4 - 2025-10-16**
 
 - Files cleaned with CleanFiles will now not be used to make pyramid tiles.
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-**Version 4.3.6 - 2025-10-27**
+**Version 4.3.5 - 2025-10-27**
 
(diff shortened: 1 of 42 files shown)


NEWS.md
```diff
@@ -1,3 +1,7 @@
+**Version 4.3.6 - 2025-10-27**
```
ernstdevreede: 4.3.5!

maartenplieger: Woops. Changed!

ernstdevreede (commented):
Version in NEWS is wrong

ernstdevreede (approved):

PR #530, opened 2025-10-28
PATCH revert replacing all spaces with underscores in dimension values.

maartenplieger:
PATCH revert replacing all spaces with underscores in dimension values.

### DIFF
--- a/adagucserverEC/CDBFileScanner.cpp
+++ b/adagucserverEC/CDBFileScanner.cpp
@@ -734,7 +734,6 @@ int CDBFileScanner::DBLoopFiles(CDataSource *dataSource, int removeNonExistingFi
                       if (dimVar->getType() == CDF_STRING) {
                         const char *str = ((char **)dimVar->data)[i];
                         uniqueKey.print("%s", str);
-                        uniqueKey.replaceSelf(" ", "_");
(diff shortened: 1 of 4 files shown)


adagucserverEC/CDBFileScanner.cpp
```diff
@@ -734,7 +734,6 @@ int CDBFileScanner::DBLoopFiles(CDataSource *dataSource, int removeNonExistingFi
                       if (dimVar->getType() == CDF_STRING) {
                         const char *str = ((char **)dimVar->data)[i];
                         uniqueKey.print("%s", str);
-                        uniqueKey.replaceSelf(" ", "_");
                         uniqueKey.replaceSelf("<", "_lt_");
```
ernstdevreede, line 737: You want to keep the replacements of <, > and = ? They are probably used nowhere but if used by accident they could cause problems?

ernstdevreede (commented):
One small question.

adagucserverEC/CDBFileScanner.cpp
maartenplieger: Jep because these endup in the GetCapabilities and cause issues

ernstdevreede (approved):

PR #554, opened 2025-11-06
Fixing barb rendering

maartenplieger:
# Changes:
- Added tests for KNMI radar loop (animated gif)
- Checked OBS dataset on geoservices with new pointrenderer implementation

# Note:
- Do not use `<Thinning` in combination with `use`
- Deprecated `Namemapping`
- Deprecated rendermethod `pointthin`
- Do not use `Point` and `Vector` together
- Rendermethod `barb` should not be used icm `Points` but is allowed for backwards compatibiity


data/config/datasets/includes/observation-styles.include.xml
```diff
@@ -0,0 +1,417 @@
...
+    <RenderMethod>point</RenderMethod>
+    <Point pointstyle="point" fillcolor="#80000000" textcolor="#000000FF" linecolor="#000000FF" discradius="0" dot="false" anglestart="0" anglestep="0" fontfile="{ADAGUC_PATH}/data/fonts/FreeSans.ttf" fontsize="8" textformat="%0.0f" plotstationid="false"/>
+  </Style>
+
+  <Style name="wwcode_transparent" title="Weather code transparent">
+    <Legend fixed="true">bluewhitered</Legend>
+    <RenderMethod>point</RenderMethod>
+    <Point pointstyle="symbol" fillcolor="#000000FF" textcolor="#FFFFFFFF" linecolor="#000000A0" discradius="16" 
+            dot="false" anglestart="0" anglestep="120" fontsize="12" textformat="%0.1f"/>
+    <SymbolInterval min="0" max="1" file="{ADAGUC_DATASET_DIR}/resources/symbols/WW-transparent/W00.png"/>
```
mgrunbauer, line 243: We don't have these symbols checked in (I guess due to size?), is that a problem? Do we have other tests which cover `SymbolInterval` usage?

adagucserverEC/CImgRenderers/CImgRenderPoints.cpp
```diff
@@ -767,10 +765,27 @@ void CImgRenderPoints::render(CImageWarper *warper, CDataSource *dataSource, CDr
...
+        }
+        if (s->FilterPoints[0]->attr.skip.empty() == false) {
+          CT::string filterPointsSkip = s->FilterPoints[0]->attr.skip.c_str();
+          std::vector<CT::string> skip = filterPointsSkip.splitToStack(",");
+          for (std::vector<CT::string>::iterator it = skip.begin(); it != skip.end(); ++it) {
+            skipPoints.insert(it->c_str());
+          }
+        }
+      }
     }
```
mgrunbauer, line 785: There is some duplicate code here, since we build the same `use/skip` sets for the Point rendering.
I guess it's okay for now, I'll try to take care of it in my follow-up refactor MR.

tests/AdagucTests/TestKMDS.py
```diff
@@ -0,0 +1,96 @@
...
+                    ]
+        for test_case in test_cases:
+            filename =test_case['filetocheck']
+            status, data, headers = AdagucTestTools().runADAGUCServer(
+                f"dataset=adaguc.test.kmds_alle_stations_10001&&SERVICE=WMS&REQUEST=GetFeatureInfo&VERSION=1.3.0&LAYERS={test_case['layers']}&QUERY_LAYERS={test_case['layers']}&CRS=EPSG%3A3857&BBOX=-141702.05839427316,6126251.383284947,1317478.0003938763,7367966.542989185&WIDTH=1550&HEIGHT=1319&I=765&J=585&FORMAT=image/gif&INFO_FORMAT=application/json&STYLES=&&time=2025-11-06T09%3A20%3A00Z", env=env, showLog=False)
+            AdagucTestTools().writetojson(self.testresultspath + filename, data.getvalue())
+            AdagucTestTools().writetojson(self.expectedoutputsspath + filename, data.getvalue())
+            self.assertEqual(status, 0)
+            self.assertEqual( json.loads(data.getvalue())[0]['units'],test_case['units'])
+            # TODO: Value does not seem to be stored in the json.
```
mgrunbauer, line 92: Regarding the TODO, is this a bug we need to fix in this MR?

tests/expectedoutputs/TestKMDS/test_kmds_alle_stations_10001_10M_animgif_ta_temperaturedisc.png
mgrunbauer, line 1: The values inside the discs are plotted behind the country borders. Is this intended?

The other png's you checked in in this MR have the same thing going on.

tests/AdagucTests/TestKMDS.py
```diff
@@ -0,0 +1,96 @@
...
+    testresultspath = "testresults/TestKMDS/"
+    expectedoutputsspath = "expectedoutputs/TestKMDS/"
+    env = {'ADAGUC_CONFIG': ADAGUC_PATH + "/data/config/adaguc.tests.dataset.xml," +
+            ADAGUC_PATH + "/data/config/datasets/adaguc.testGeoJSONReader_time.xml"}
+
+    AdagucTestTools().mkdir_p(testresultspath)
+
+
+
+    def test_kmds_alle_stations_10001_getmap_requests(self):
```
mgrunbauer, line 27: Note, test cases like these can be rewritten a bit nicer (IMO) without using for loops in the test case itself by using `pytest.mark.parametrize`. Pytest will see them as separate test cases, so you can also run an individual case (e.g. just run `barballpoint`).

However, we can't switch to it right now because we're still using `unittest.TestCase`.

Here's an example of how it could be done in the future with parametrize:
```
    @pytest.mark.parametrize("filetocheck,layers,styles", [
        ("test_kmds_alle_stations_10001_10Mwind_barballpoint.png","10M%2Fwind", "barball%2Fpoint"),
        ("test_kmds_alle_stations_10001_10Mwind_barbthinnedpoint.png", "10M%2Fwind", "barbthinned%2Fpoint"),
        # more test cases here
    ])
    def test_kmds_alle_stations_10001_getmap_requests(self, filetocheck: str, layers: str, styles: str):
        # you have access to the individual parameters here:
        print(filetocheck, layers, styles)
        pass
```

data/config/datasets/includes/observation-styles.include.xml
maartenplieger: No we don't have tests with symbolinterval...

tests/AdagucTests/TestKMDS.py
maartenplieger: Should be fixed in this bug report: https://github.com/KNMI/adaguc-server/issues/559

is already the case in master

tests/expectedoutputs/TestKMDS/test_kmds_alle_stations_10001_10M_animgif_ta_temperaturedisc.png
maartenplieger: Yes that is intended. You can see them over the baselayer and under the overlayer. If you plot them on top you can not see the countryborders anymore.

tests/AdagucTests/TestKMDS.py
maartenplieger: Would be nice to do in a new MR!

data/config/datasets/includes/observation-styles.include.xml
maartenplieger: Added one tests with cloudcover which uses SymbolInterval.

mgrunbauer (approved):

PR #563, opened 2025-11-11
Fix units

ernstdevreede:
This MR fixes the ability to override units and standard_name fields for <Variable> elements in a layer.

It also fixes the handling of NC_STRING attributes by ADAGUC.


adagucserverEC/CDataSource.cpp
```diff
@@ -998,6 +998,28 @@ int CDataSource::attachCDFObject(CDFObject *cdfObject, bool dataSourceOwnsDataOb
...
+    if (var != nullptr) {
+
+      // Set long_name
+      if (!cfgVar->attr.long_name.empty()) {
+        var->setAttributeText("long_name", cfgVar->attr.long_name);
+      }
+
+      // Set units
+      if (!cfgVar->attr.units.empty()) {
+        CDBDebug("Trying to set unit %s", cfgVar->attr.units.c_str());
```
maartenplieger: Can be removed

ernstdevreede: Removed

maartenplieger (approved):

PR #567, opened 2025-11-12
Refactoring to make code simpler and more C++ like for geoparams

### LINKED ISSUE #568: Refactoring to make code simpler and more C++ like for geoparams
Refactoring to make code simpler and more C++ like for geoparams

See https://github.com/KNMI/adaguc-server/pull/567

maartenplieger:
Only refactoring, no changed functionality

1. Use of f8box in CGeoParams
2. CGeoParams is now a struct which can be copied by assign operator
3. Avoid using CGeoParams* where possible and do not use new to allocate it.
4. Geo is renamed to geoParams
5. GDWDrawFunctionSettings is moved to its own file adagucserverEC/GenericDataWarper/GDWDrawFunctionSettings.h
6. Avoid using classes with new and pointers where structs could be used.

These are a huge amount of changes but functionality should be unchanged.

Closes https://github.com/KNMI/adaguc-server/issues/568


adagucserverEC/Types/GeoParameters.h
```diff
@@ -0,0 +1,79 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ ******************************************************************************/
+
+#ifndef CGeoParams_H
+#define CGeoParams_H
+#include "CTypes.h"
+#include <math.h>
+#include <map>
```
mgrunbauer, line 30: I don't think you need `map` and `CKeyValuePair` here?

I am not sure about the others.

adagucserverEC/utils/CRequestUtils.cpp
```diff
@@ -23,27 +23,27 @@ std::tuple<int, std::array<double, 4>> findBBoxForDataSource(std::vector<CDataSo
       //   try {
```
belentorrente: Will we ever use this code again? Maybe a good opportunity to remove it. We can always recover it going back in the commit history.

adagucserverEC/utils/CRequestUtils.cpp
```diff
@@ -23,27 +23,27 @@ std::tuple<int, std::array<double, 4>> findBBoxForDataSource(std::vector<CDataSo
...
       if (status != 0) {
         warper.closereproj();
         CDBDebug("Unable to initialize projection ");
       }
       warper.findExtent(dataSources[d], dfBBOX);
       warper.closereproj();
-      CDBDebug("Found bbox %s %f %f %f %f", dataSources[0]->srvParams->Geo->CRS.c_str(), dfBBOX[0], dfBBOX[1], dfBBOX[2], dfBBOX[3]);
-      std::array<double, 4> box = {dfBBOX[0], dfBBOX[1], dfBBOX[2], dfBBOX[3]};
+      CDBDebug("Found bbox %s %f %f %f %f", dataSources[0]->srvParams->geoParams.crs.c_str(), dfBBOX[0], dfBBOX[1], dfBBOX[2], dfBBOX[3]);
+      f8box box = {dfBBOX[0], dfBBOX[1], dfBBOX[2], dfBBOX[3]};
```
mgrunbauer, line 31: I discovered we also have `BBOX.h`, which seems very similar to `f8box`.

Could/should we use the new `f8box` here? Maybe we can pass `f8box` to `warper.findExtent` directly, then we don't need the extra `double[4]`.

adagucserverEC/CConvertLatLonBndsData.cpp
```diff
@@ -113,26 +113,24 @@ int CConvertLatLonBnds::convertLatLonBndsData(CDataSource *dataSource, int mode)
...
   }
 
   // Width needs to be at least 2 in this case.
   if (dataSource->dWidth == 1) dataSource->dWidth = 2;
   if (dataSource->dHeight == 1) dataSource->dHeight = 2;
-  double cellSizeX = (dataSource->srvParams->Geo->dfBBOX[2] - dataSource->srvParams->Geo->dfBBOX[0]) / double(dataSource->dWidth);
-  double cellSizeY = (dataSource->srvParams->Geo->dfBBOX[3] - dataSource->srvParams->Geo->dfBBOX[1]) / double(dataSource->dHeight);
-  double offsetX = dataSource->srvParams->Geo->dfBBOX[0];
-  double offsetY = dataSource->srvParams->Geo->dfBBOX[1];
+  double cellSizeX = (dataSource->srvParams->geoParams.bbox.right - dataSource->srvParams->geoParams.bbox.left) / double(dataSource->dWidth);
```
belentorrente: For cellSizeX and cellSizeY, these operations are so frequent, maybe you can add them to the `GeoParameters` as class methods, instead of calculating them all the time in the code?

adagucserverEC/utils/CRequestUtils.cpp
maartenplieger: Wrote https://github.com/KNMI/adaguc-server/issues/570

adagucserverEC/utils/CRequestUtils.cpp
maartenplieger: Good point, done!

adagucserverEC/CConvertLatLonBndsData.cpp
maartenplieger: At least span is now used. It is a bit difficult to get the dataSource->dWidth towards the bbox.

adagucserverEC/Types/GeoParameters.h
maartenplieger: Removed

mgrunbauer (approved):

belentorrente (approved):

PR #569, opened 2025-11-12
Support EDR calls with hidden dimensions

### LINKED ISSUE #565: EDR does not work for WMS Layer with hidden=true
A hidden layer configured like this:
```
<Layer hidden="true">
    ...
</Layer>
```

should not be advertised in WMS GetCapabilities, but EDR calls should work. Currently we don't add hidden layers to GetMetadata at all. We need to add layers to the metadata even if they are hidden.

mgrunbauer:
- Hidden layer will now end up in metadata API and will work for all EDR calls. hidden=true only results in GetCapabilities not including this layer.
- EDR will skip `Temporal` if the metadata layer has no time dimension. This happens with things like hidden base layers/geojson.
- Store some test result jsons as files.
- Minor refactor, using `for (auto` in for loops.

Fixes https://github.com/KNMI/adaguc-server/issues/565.


adagucserverEC/CRequest.cpp
```diff
@@ -2900,77 +2900,75 @@ int CRequest::handleGetMapRequest(CDataSource *firstDataSource) {
...
-        driverName.copy(srvParam->cfg->WCS[0]->WCSFormat[i]->attr.driver.c_str());
-        break;
-      }
-    }
-    if (driverName.equals("ADAGUCNetCDF")) {
-      CDBDebug("Creating CNetCDFDataWriter");
-      wcsWriter = new CNetCDFDataWriter();
+  if (srvParam->requestType != REQUEST_WCS_GETCOVERAGE) {
+    return status;
+  }
```
mgrunbauer, line 2910: It looks like a lot of changes, but I've only inverted this if statement and changed `for (size_t...` to `for (auto...`.

adagucserverEC/CImageDataWriter.cpp
```diff
@@ -673,10 +672,12 @@ int CImageDataWriter::getFeatureInfo(std::vector<CDataSource *> &dataSources, in
   // Create a new getFeatureInfoResult object and push it into the vector.
   int status = 0;
   isProfileData = false;
-  for (size_t d = 0; d < dataSources.size(); d++) {
-    if (dataSources[d]->cfgLayer->attr.hidden.equals("true")) {
-      continue;
+  for (auto dataSource : dataSources) {
```
mgrunbauer, line 675: Instead of passing a `dataSources[d]` everywhere, we just use a `auto dataSource` directly. 

Small changes were required in `CMakeEProfile::MakeEProfile` and `CMakeJSONTimeSeries::MakeJSONTimeSeries` so that they only accept a single `dataSource` instead of a vector and a index.

mgrunbauer:
Add new test specifically for hidden layer

maartenplieger (approved):

PR #592, opened 2026-01-12
Support pointsfromgrid on windspeed components

maartenplieger:
**Version 6.2.1 - 2026-01-12**

- pointsfromgrid can now also be used to plot text values on windspeed and winddirection components


adagucserverEC/CImgRenderers/CImgRenderPoints.cpp
```diff
@@ -407,8 +407,11 @@ void renderSinglePoints(std::vector<size_t> thinnedPointIndexList, CDataSource *
 
   float fillValueObjectOne = dataSource->getDataObject(0)->hasNodataValue ? dataSource->getDataObject(0)->dfNodataValue : NAN;
   for (size_t dataObjectIndex = 0; dataObjectIndex < dataSource->getNumDataObjects(); dataObjectIndex++) {
-    std::vector<PointDVWithLatLon> *pts = &dataSource->getDataObject(dataObjectIndex)->points;
-
+    auto dataObject = dataSource->getDataObject(dataObjectIndex);
+    if (dataObject == nullptr) continue;
+    std::vector<PointDVWithLatLon> *pts = &dataObject->points;
+    if (pts == nullptr) continue;
```
mgrunbauer, line 413: These checks make sense. I think they are also required in the other render methods in this file. For example, in `renderSingleVolumes` we do the following:
```cpp
std::vector<PointDVWithLatLon> *pts = &dataSource->getDataObject(dataObjectIndex)->points;
```
and in theory `dataSource->getDataObject(dataObjectIndex)` could also be `null` here?

Same for `renderSingleDiscs`, `renderSingleDot`, `renderSingleSymbols` and maybe also `renderVectorPoints`.

maartenplieger: Added checks for those as well.

mgrunbauer (dismissed):

mgrunbauer (dismissed):

mgrunbauer (approved):

PR #612, opened 2026-01-29
New add_dataobject datapostproc to support wave direction

### LINKED ISSUE #589: Support plotting wave direction as vector
Plot a vector based on a NetCDF parameter with 0-360 degree value.

mgrunbauer:
Fixes: https://github.com/KNMI/adaguc-server/issues/589

This PR adds support for the following use-cases:

1. Visualizing direction in degrees as a vector, using a new `add_dataobject` datapostproc.

2. Visualizing direction+speed (or something else) as a vector, from different netcdf files, by using `include_layer` datapostproc.

**1: Direction only**
To draw a vector you need a "strength" and a "direction" component.

To visualize _just_ direction in degrees as a vector, you need to create a dummy strength component. This PR adds the `add_dataobject` datapostproc. Can be configured like this:
```xml
<DataPostProc algorithm="add_dataobject" name="dummy-strength" a="1" />
```
Combining this with `pointsfromgrid` datapostproc and `<Vector>` allows you to draw just the direction as a vector.

**2: Direction + speed from two files**
Add an example test case that demonstrates how to load variables from two different NetCDF files and combine them in a single layer


data/config/datasets/adaguc.tests.vectorrendering.xml
```diff
@@ -118,4 +118,60 @@
     <RenderSettings drawgrid="false" />
     <DataPostProc algorithm="pointsfromgrid" select="speed_component,direction_component" a="150"/>
   </Style>
+
+  <!-- wave plotting -->
+  <Style name="wave_direction">
+    <DataPostProc algorithm="add_dataobject" name="fake-strength" a="1" />
```
maartenplieger: - [x] Can you add documentation to where the other datapostprocessors are described?
- [x] Can you also write a small markdown file, like a short tutorial on how to use this?
- [x] Can you add something to the news describing this new feature?

maartenplieger (approved):

PR #615, opened 2026-02-02
No error macros and format checking for debug statements.

maartenplieger:
- Removed error macros.
- Removed className in debugger.
- Removed extern and direct initialization where possible
- Added format checker to debug macro's
- Added standalone ctprintf function


adagucserverEC/CDataSource.h
```diff
@@ -25,21 +25,17 @@
...
+#include "CCDFObject.h"
 #include "CStyleConfiguration.h"
+#include "Types/CPointTypes.h"
+#include "COGCDims.h"
+#include "Types/GeoParameters.h"
+#include "CServerParams.h"
 
-#include "CGeoJSONData.h"
+// Forward declaration
+class CStyleConfiguration;
```
mgrunbauer: `clang` complains because this is actually a `struct` and not a `class`.

I think this is correct?
```c++
// Forward declaration
struct CStyleConfiguration;
```

mgrunbauer:
Now compiles with `clang`, but a C++ test fails:

```
2: Test command: /Users/myuser/src/adaguc-server-review/bin/adagucserverEC/testadagucserver
2: Working Directory: /Users/myuser/src/adaguc-server-review/bin
2: Test timeout computed to be: 10000000
2: [E:000:pid97430: adagucserverEC/testadagucserver.cpp:48]                        linearTransform [-1.000000] != [0.000000] wrong result
2: libc++abi: terminating due to uncaught exception of type int
```

mgrunbauer (approved):

mgrunbauer:
Works with a release build. Debug build still produces errors on clang but that's an issue for another day :)

PR #621, opened 2026-02-02
Parametrize tests

### LINKED ISSUE #560: Refactor tests using pytest parametrize
We have multiple test cases combined in a single test in `TestKMDS.py`. We use a for loop to handle all the cases, however it can be a bit tricky to debug _some_ of these test cases.

We can improve this by using `pytest.mark.parametrize`, the code would be something like this:

```python
    @pytest.mark.parametrize("filetocheck,layers,styles", [
        ("test_kmds_alle_stations_10001_10Mwind_barballpoint.png","10M%2Fwind", "barball%2Fpoint"),
        ("test_kmds_alle_stations_10001_10Mwind_barbthinnedpoint.png", "10M%2Fwind", "barbthinned%2Fpoint"),
        # more test cases here
    ])
    def test_kmds_alle_stations_10001_getmap_requests(self, filetocheck: str, layers: str, styles: str):
        # you have access to the individual parameters here:
        print(filetocheck, layers, styles)
        pass
```

Pytest will see these as separate test cases, so you can just run an individual case (e.g. just run barballpoint).

However, we can't switch to parametrize right now because we're still using unittest.TestCase as a base class for our tests.
`unittest.TestCase` gives us things like `assertEquals`, however we can replace these with either pytest fixtures in `conftest.py` or by just doing `assert x == y` for certain things.

IMO we should move towards `pytest`, note that this is our test runner since: https://github.com/KNMI/adaguc-server/pull/516.

mgrunbauer:
Fixes: https://github.com/KNMI/adaguc-server/issues/560

Parametrize tests using pytest.

Because `unittest.TestCase` and pytest `parametrize` do not work together, I've converted all the `self.assertTrue` to regular python `assert x == y` calls.

I've added the following helper methods to `conftest.py`:
```python
env = make_adaguc_env("/path/to/testcase_config.xml")
update_db(env)
```

- `make_adaguc_env` gives `{"ADAGUC_CONFIG": "/path/to/default_config.xml,/path/to/testcase_config.xml"}`.
- `update_db` calls `--updatedb` once per given dataset/env. This means we don't needlessly update the database.

If we're happy with the helper functions and the new asserts, we could refactor all other test files (in a new PR). I think it will reduce testing boiler plate by quite a bit.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -90,16 +90,18 @@ def make_adaguc_env(dataset: str = "") -> dict:
     return {"ADAGUC_CONFIG": config}
 
 
-def update_db(adaguc_env: dict):
+def update_db(adaguc_env: dict, force_update: bool = False):
     """
-    Runs `--updatedb` if it hasn't been ran before. When executed multiple times with the same env, no extra calls will take place.
+    Run `--updatedb` if it has not been run before for the given environment. When executed multiple times
+    with the same environment, no additional `--updatedb` calls will be made, unless `force_update` is enabled.
 
-    Expects an `adaguc_env`, which is a dictionary with `ADAGUC_CONFIG` as key.
+    - `adaguc_env`: Environment configuration with `ADAGUC_CONFIG` as key.
+    - `force_update`: Always execute `force_update`, even if it has run before.
     """
 
     adaguc_config = adaguc_env.get("ADAGUC_CONFIG")
 
-    if adaguc_config in DATASETS_LOADED:
+    if not force_update and adaguc_config in DATASETS_LOADED:
         return
 
     from adaguc.AdagucTestTools import AdagucTestTools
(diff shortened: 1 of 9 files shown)


tests/conftest.py
```diff
@@ -68,6 +68,47 @@ def ensure_adaguc_test_db():
...
+
+    config = f"{ADAGUC_PATH}/data/config/adaguc.tests.dataset.xml"
+
+    if dataset:
+        config = f"{config},{dataset}"
+
+    return {"ADAGUC_CONFIG": config}
+
+
+def update_db(adaguc_env: dict):
```
maartenplieger: Can you add a force property so it will scan anyway?

mgrunbauer: Done!

maartenplieger (approved):

PR #642, opened 2026-02-13
Added stride option to solt to make it render faster

maartenplieger:
Added striding option so it does not need to be calculated for each pixel.

image


adagucserverEC/CDataPostProcessors/CDataPostProcessor_SolarTerminator.cpp
```diff
@@ -46,23 +49,34 @@ int CDPPSolarTerminator::execute(CServerConfig::XMLE_DataPostProc *proc, CDataSo
     // Copy bounding box of screen
     dataSource->setGeo(dataSource->srvParams->geoParams);
 
+    if (dataSource->srvParams->requestType != REQUEST_WMS_GETFEATUREINFO) {
+      dataSource->dWidth = dataSource->srvParams->geoParams.width / 16;
```
belentorrente: What is the meaning of the number 16 here?

adagucserverEC/CImgWarpGeneric/CImgWarpGeneric.cpp
```diff
@@ -32,6 +32,7 @@
 CColor cblack = CColor(0, 0, 0, 255);
 CColor cblue = CColor(0, 0, 255, 255);
 
+MemoizitationForDeterminePixelColorFromValue memo;
```
belentorrente: Typo - It's probably "memoization"

adagucserverEC/LayerTypeLiveUpdate/LayerTypeLiveUpdate.cpp
```diff
@@ -174,19 +142,23 @@ int layerTypeLiveUpdateRenderIntoImageDataWriter(CDataSource *dataSource, CServe
   // Covers case of Solar Terminator
   CImageDataWriter imageDataWriter;
   int status = imageDataWriter.init(srvParam, dataSource, dataSource->getNumTimeSteps());
-  CDBDebug("Init imageDataWriter status %d", status);
+  if (verbose) {
```
belentorrente, line 154: Maybe we can add a symbol to do live update debug info printing?

adagucserverEC/CGenericDataWarper.cpp
```diff
@@ -444,6 +444,7 @@ template <typename T>
 int GenericDataWarper::render(CImageWarper *warper, void *_sourceData, GeoParameters sourceGeoParams, GeoParameters destGeoParams,
                               const std::function<void(int, int, T, GDWState &warperState)> &drawFunction) {
 
+  bool verbose = false;
```
belentorrente, line 447: Can this be part of the generic data warper debug info? Printed if the debug symbol is defined.

doc/tutorials/Configure_solar_terminator.md
```diff
@@ -48,6 +48,9 @@ Feel free to customize the style to your preference, such as adding a fill color
 
 The `offset` attribute is optional, with the default being one year. The time range that will be made available is `(now-offset, now+offset)`.
 
+
+The `stride` attribute is optional, it determines how many pixels in the getmap request should be grouped before doing a new calculation
```
belentorrente, line 52: I did not fully understand the concept of stride. If it was 2, would you group 2x2 pixels?

tests/AdagucTests/TestWMS.py
```diff
@@ -2440,14 +2437,13 @@ def test_WMSGetLegendGraphic_SolarTerminator_continuous(self):
             AdagucTestTools().readfromfile(self.expectedoutputsspath + filename),
         )
 
-
     def test_WMSGetFeatureInfo_SolarTerminator(self):
         # Testing the solar terminator feature info
-        env = make_adaguc_env("{ADAGUC_PATH}/data/config/datasets/adaguc_tests_solarterminator.xml")
+        env = make_adaguc_env("{ADAGUC_PATH}/data/confi`g/datasets/adaguc_tests_solarterminator.xml")
```
belentorrente: this is breaking the test

adagucserverEC/CDataPostProcessors/CDataPostProcessor_SolarTerminator.cpp
```diff
@@ -17,6 +17,9 @@
 /*      CDPPSolarTerminator  */
 /************************/
 
+#define MAX_SOLT_WIDTH 1024
```
belentorrente: I think you ended up not needing this? I don't see it in the diff at first glance

adagucserverEC/CDataPostProcessors/CDataPostProcessor_SolarTerminator.cpp
maartenplieger: Changed to DEFAULT_STRIDING

adagucserverEC/CImgWarpGeneric/CImgWarpGeneric.cpp
maartenplieger: Changed!

doc/tutorials/Configure_solar_terminator.md
maartenplieger: Yes. It is currenlty rendering the grid at a lower resolution than the screen resolution. If you would say striding= 2, the width and height would be half that of the screen resolution.

tests/AdagucTests/TestWMS.py
maartenplieger: Fixed.

adagucserverEC/CDataPostProcessors/CDataPostProcessor_SolarTerminator.cpp
maartenplieger: Removed!

adagucserverEC/LayerTypeLiveUpdate/LayerTypeLiveUpdate.cpp
maartenplieger: I think we will only set this to true manually when developing new features in this area. It is similar as the debug macro definitions in Definitions.h. Both doese `#ifdef` `#endif` statements make the code very hard to read.

I would suggest to think on how we handle this more generic. For now I think it is OK in this PR.

belentorrente: I think we need something like the Python logger, maybe by library? And yes, it can be its own ticket.

doc/tutorials/Configure_solar_terminator.md
belentorrente: This is very interesting to use with prefetching. You can get many timesteps in a short(er) time

belentorrente (approved):

PR #606, opened 2026-01-27
Fix nan points

ernstdevreede:
This PR fixes handling of NAN point values. In the old code NAN values resulted in automatic plotting of the station_id, which could cause confusion.
Also changes the way multi-parameter point data is visualised: angelstep and aglestart atgtributes have been removed (were used for plotting multiple  values around a point). Multiple parameters are now visualised in a vertical stack, centered on the point data's coordinates.

PR #656, opened 2026-03-02
Fixed issue where geojson points can not be rendered if values have nan

maartenplieger:
- KMDS files where stations have NAN values do not show anything, as expected.
- Mistposten file in GeoJSON now renders as expected
- Detection of extenstions has been refactored
- Added more utility functions to CT::string


loescornelis (dismissed):

adagucserverEC/CImgRenderers/CImgRenderPoints.cpp
```diff
@@ -437,10 +437,14 @@ void renderSinglePoints(std::vector<size_t> thinnedPointIndexList, CDataSource *
 
       int y = calculateYForDataobject(dataSource, station_y, pointStyle.textRadius, dataObjectIndex);
 
-      if (dataType == CDF_STRING) {
+      // If value is nan, we can still try to print a text from the paramlist.
+      if (dataType == CDF_STRING || std::isnan(value)) {
```
ernstdevreede: This change undoes the effect of the recent fix-nan-points PR (https://github.com/KNMI/adaguc-server/pull/606/).
The effect is that now ADAGUC plots the station name again (or any other first value of paramList) when data is nan (which is sometimes the case for KMDS data, but always the case for GeoJSON).

data/config/datasets/adaguc.tests.geojsonpoint.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<Configuration>
+
+
+  <Style name="stations_mistpost">
+    <Legend fixed="true">bluewhitered</Legend>
+    <Thinning radius="40"/>
```
ernstdevreede: I would not use thinning for this data, because you probably don't want to miss a mistpost.

ernstdevreede (changes requested):
I think we have to look at this solution as it does not work OK everywhere.

adagucserverEC/CImgRenderers/CImgRenderPoints.cpp
ernstdevreede: I believe the problem is that although in the XML we have configured `name` as the Variable the actual value of `name` does not end up in the value field.

maartenplieger: The testcases for #606 still work, so I wonder if this is really the case.

ernstdevreede: That is my fault: I overlooked the fact that the Actuele10mindataKNMIstations_*.nc files don't use NAN's to indicate missing values like the new KMDS files do. The Actuele10mindataKNMIstations_*nc files use _FillValue attributes for that.
So #606 was not fully covered by test cases.
Have to fix that.

maartenplieger: Thanks Ernst. Will have a look on how can make both cases work as we want.

maartenplieger: Added ADAGUC_ORGPOINT_TYPE attribute in the GeoJSON reader. Now all works!

data/config/datasets/adaguc.tests.geojsonpoint.xml
maartenplieger: Removed.

ernstdevreede (approved):

PR #683, opened 2026-04-13
Replaced CAIRO_LINE_JOIN_MITER with BEVEL

loescornelis:
https://www.cairographics.org/manual/cairo-cairo-t.html#cairo-line-join-t 

This creates less "harsh" spikes


adagucserverEC/CCairoPlotter.cpp
```diff
@@ -562,12 +563,12 @@ void CCairoPlotter::lineTo(float x1, float y1, float width) {
 }
 
 void CCairoPlotter::endLine() {
-  cairo_set_line_join(cr, CAIRO_LINE_JOIN_MITER);
```
maartenplieger, line 565: Shall we just do this? Can you update the tests?

loescornelis: Yes done! Looks a lot better in my opinion.

maartenplieger (approved):

PR #685, opened 2026-04-14
Improve error handling

maartenplieger:
- Error handling is refactored
- JSON error handling is improved


adagucserverEC/CDataPostProcessors/CDataPostProcessor_IncludeLayer.cpp
```diff
@@ -37,7 +37,7 @@ int CDPPIncludeLayer::setDimsForNewDataSource(CServerConfig::XMLE_DataPostProc *
     if (CRequest::setDimValuesForDataSource(dataSourceToInclude, dataSource->srvParams) == 0) {
       dataIsFound = true;
     }
-  } catch (ServiceExceptionCode e) {
+  } catch (ServiceExceptionType e) {
```
mgrunbauer, line 40: Just curious about this bit:

- if we're unable to set dims, we catch the exception
- we don't do anything in the catch
- we debug print "no data available"

Should this maybe become a `CDBWarning`?
Alternatively raise exception since we really cannot find data that is expected... but maybe this happens often and you don't want to error on this.

mgrunbauer (dismissed):

adagucserverEC/CDataPostProcessors/CDataPostProcessor_IncludeLayer.cpp
maartenplieger: Added a warning statement instead.

mgrunbauer (approved):

PR #690, opened 2026-04-15
Brotli encoding is now done only for certain Content-Types.

maartenplieger:
- Brotli encoding is now done only for certain Content-Types.


python/python_fastapi_server/routers/caching_middleware.py
```diff
@@ -15,6 +15,9 @@
 
 MAX_SIZE_FOR_CACHING = 10_000_000
 
+BR_COMPRESS_CONTENT_TYPES = ["application/json", "text/xml", "text/html", "text/plain", "image/png32"]
```
mgrunbauer: You can make this a set. If you then check like this:

```
for k, v in headers.items():
    if k.lower() == "content-type":
            return v in BR_COMPRESS_SET
    return False
```

You have a quick lookup without having to see _all_ the headers.

mgrunbauer: Why should we still compress png32, isn't that also already compressed pixeldata?

ernstdevreede (approved):

PR #696, opened 2026-04-17
No extra bytesio

mgrunbauer:
- `CGIRunner.py` previously wrote adaguc output to a BytesIO buffer, which was then immediately read using `getvalue()` in the next function. This MR removes the extra buffering step. This means we don't need to call `getvalue()` anymore. All tests were touched because they all did `getvalue()`.

- Small python improvements, added some typehints

- I moved some tests from TestWMS:
    - Solar Terminator to `TestWMSLiveUpdate`
    - The new points and vecors tests (added after refactoring `CImgRenderPoints`) to `TestPointAndVectors`
    - Tests which only call `GetLegendGraphic` to `TestWMSGetLegendGraphic`.
    - All these moved tests are now using `make_adaguc_env` and `update_db(env)`


maartenplieger (approved):

maartenplieger:
Looks good to me..

PR #698, opened 2026-05-04
All other dimensions like groundlevel or elevation now also work with getmetadata and a specific reference time

maartenplieger:
All other dimensions like groundlevel or elevation now also work with getmetadata and a specific reference time

### DIFF
--- a/tests/AdagucTests/TestMetadataRequest.py
+++ b/tests/AdagucTests/TestMetadataRequest.py
@@ -52,7 +52,7 @@ class TestMetadataRequest:
 
     def test_timeseries_adaguc_tests_arcus_uwcw_air_temperature_hagl_non_existing_referencetime(self):
         """
-        This checks for correct behavior when passing a invalid reference time, a 404 should be returned.
+        This checks for correct behavior when passing a invalid reference time, a status OK should be returned with empty dims.
         """
         env = make_adaguc_env("adaguc.tests.arcus_uwcw", self.testresultspath, self.expectedoutputsspath)
         update_db(env, True)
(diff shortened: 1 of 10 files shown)


tests/AdagucTests/TestMetadataRequest.py
```diff
@@ -52,26 +52,27 @@ def test_timeseries_adaguc_tests_arcus_uwcw_air_temperature_hagl_specific_query(
 
     def test_timeseries_adaguc_tests_arcus_uwcw_air_temperature_hagl_non_existing_referencetime(self):
         """
-        This checks for correct behavior when passing a invalid reference time, a 404 should be returned.
+        This checks for correct behavior when passing a invalid reference time, a status OK should be returned with empty dims.
```
mgrunbauer, line 55: This test asserts that when we query a wrong dim_reference_time in the metadata table, adaguc returns a 200 OK.

Isn't a wrong dim_reference_time a good reason to through a non-200 status code?
We also have a test case in `test_ogc_api_edr.py` line 726 which asserts that a call to `/edr/collections/adaguc.tests.members.mycollection/instances/12345` results in a HTTP 404 with message `Incorrect instance 12345 for collection adaguc.tests.members.mycollection`. I feel like these two things are clashing a bit now?

maartenplieger: Some layers return results and others do not.

To be able to list all layers, we now have to prevent returning an error for a specific layer. 

Now we always return data. Even if the reference time is wrong for one of the layers..

mgrunbauer (approved):

PR #702, opened 2026-05-13
Update documentation and remove unused environment variable

mgrunbauer:
- Updated profiling documentation. New docs install debug symbols for adaguc-server dependencies, and use the correct profiling format. This works well inside docker.

- Formatted environment variable documentation in a table. I did this earlier in my fork branch to keep an overview of environment variables, I think it fits better in this (update docs) PR.

- Removed `ADAGUC_ERRORFILE` environment variable. It appears we're not using this anymore. Because of this I needed to save and format `ADAGUCWCS.py`.


doc/developing/profiling.md
```diff
@@ -2,59 +2,70 @@
...
+```bash
+docker run --rm --privileged --net=host --pid=host alpine \
+  sysctl -w kernel.perf_event_paranoid=-1 \
+  kernel.kptr_restrict=0
+```
 
+Edit `./Docker/docker-compose.yml` to create profiling build and allow capturing perf events in the container:
 ```yml
   adaguc-server:
+    build:
```
maartenplieger, line 24: Can you either add to the docker compose or in these instructions:

```
     extra_hosts:
      - "host.docker.internal:host-gateway" 
   ```

In the build section of adaguc-server?

File: doc/developing/profiling.md [hunk 2]
```diff
@@ -79,64 +90,39 @@ If you execute `adagucserver` manually, it will write the (binary) response to t
...
 
 ```bash
-perf record -o perf.out -F1000 -g --call-graph dwarf ./bin/adagucserver
-```
-
-
-You need to "post process" the outputted file. This step needs to happen inside the docker container as well:
-```bash
-perf script -i perf.out | c++filt > perf.data
+perf record -o perf.data -F max -g --call-graph fp ./bin/adagucserver
```
maartenplieger, line 100: Can you also say that dwarf is an alternative profiler thingy?

maartenplieger (approved):

PR #721, opened 2026-06-25
Refactor png writing code

mgrunbauer:
We discovered during profiling that `InsertTree` is a significant part of creating a 8-bits PNG.

I refactored the PNG writing code and made the following changes:
- Everything relating to writing PNG moved from `CCairoPlotter.cpp` to `png_write.cpp`.
- Removed 32 bit PNG (not used)
- Refactored 1 method into 5 separate methods: `writePng`, `prepare8Bpp`, `write8BppPayload`, `prepare24Bpp` and `write24BppPayload`.
- Still produces identical images according to the tests.

Specifically for PNG performance, I changed the following:
- Instead of inserting individual pixels into the octree using `InsertTree`, group identical sequential colors together and add them once. This keeps palette generation identical, while reducing octree traversal.
- Instead of looking up every pixel in the octree (which means traversing the octree until the color was found), use a cache of colors.

Also performance for `CCairoPlotter::CCairoPlotter`:
- Instead of writing per color channel (8 bits), it's quicker to write per 32 bits.

Also fixed some warnings with new clang version.


adagucserverEC/CCairoPlotter.cpp
```diff
@@ -179,16 +180,17 @@ CCairoPlotter::CCairoPlotter(int width, int height, float fontSize, std::string
...
-      ARGBByteBuffer[j * 4 + 2] = r;
-      ARGBByteBuffer[j * 4 + 1] = g;
-      ARGBByteBuffer[j * 4 + 0] = b;
+    uint32_t pixel = (a << 24) | (r << 16) | (g << 8) | b;
+
+    auto *quadBuffer = reinterpret_cast<uint32_t *>(ARGBByteBuffer);
+    size_t pixelCount = bufferSize / 4;
+
+    for (size_t i = 0; i < pixelCount; ++i) {
+      quadBuffer[i] = pixel;
```
maartenplieger: You can  probably do:

 std::memset(quadBuffer, pixel, pixelCount);

adagucserverEC/COctTreeColorQuantizer.cpp
```diff
@@ -16,24 +17,29 @@ static uint glbTotalLeaves = 0;
...
+    (*tree)->realalphasum += ulong(color->realalpha) * count;
   } else {
-    InsertTree(&((*tree)->child[LEVEL(color, TREEDEPTH - depth)]), color, depth + 1);
+    InsertTreeCount(&((*tree)->child[LEVEL(color, TREEDEPTH - depth)]), color, depth + 1, count);
   }
 }
+
+// InsertTree -- Insert a color into the octree
+//
+void InsertTree(OctreeType **tree, RGBType *color, uint depth) { InsertTreeCount(tree, color, depth, 1); }
```
maartenplieger: Is this still used?

adagucserverEC/write_png.cpp
```diff
@@ -0,0 +1,338 @@
...
+  }
+
+  if (!insertedVisiblePixel) {
+    color.r = 0;
+    color.g = 0;
+    color.b = 0;
+    color.realblue = 0;
+    color.realalpha = 0;
+
+    InsertTree(&tree, &color, -1);
```
maartenplieger: Maybe use InsertTreeCount(&tree, &color, -1, 1);  instead?

adagucserverEC/CCairoPlotter.cpp
mgrunbauer: Funny thing, `std::memset` uses 8-bits, even when you feed it a 32-bits type...
But we can use `std::fill_n` for both cases.

maartenplieger (approved):

PR #728, opened 2026-07-01
Refactor cdbstore

maartenplieger:
- Refactored CT::string to std::string in CDBStore
- Removed unnecessary c_str().

### DIFF
--- a/adagucserverEC/CDBAdapterPostgreSQL.cpp
+++ b/adagucserverEC/CDBAdapterPostgreSQL.cpp
@@ -100,7 +100,7 @@ CT::string CDBAdapterPostgreSQL::getDimValueForFileName(const char *filename, co
   CT::string query;
   query.print("select * from %s where path = '%s' limit 1", table, filename);
   CDBStore::Store *store = DB->queryToStore(query.c_str());
-  if (store == NULL || store->size() == 0) {
+  if (store == NULL || store->records.size() == 0) {
     setExceptionType(InvalidDimensionValue);
     CDBError("Invalid filename value for  %s", filename);
     CDBError("query failed");
@@ -109,7 +109,7 @@ CT::string CDBAdapterPostgreSQL::getDimValueForFileName(const char *filename, co
 #ifdef MEASURETIME
   StopWatch_Stop("<CDBAdapterPostgreSQL::getDimValueForFileName");
 #endif
-  CT::string dimValue = *store->getRecord(0)->get(1);
+  CT::string dimValue = store->records[0].get(1);
   delete store;
   return dimValue;
 };
@@ -495,7 +495,7 @@ CDBStore::Store *CDBAdapterPostgreSQL::getFilesAndIndicesForDimensions(CDataSour
     return NULL;
   }
   if (raiseExceptionWhenOverLimit && limit > 0) {
-    if (int(store->size()) > limit) {
+    if (int(store->records.size()) > limit) {
       delete store;
       CDBError("Requested data exceeded maxquerylimit=%d", limit);
       throw UnprocessableEntity;
@@ -544,18 +544,21 @@ int CDBAdapterPostgreSQL::autoUpdateAndScanDimensionTables(CDataSource *dataSour
     if (store == NULL) {
       tableNotFound = true;
       CDBDebug("No table found for dimension %s", dimName.c_str());
+    } else if (store->records.size() == 0) {
+      tableNotFound = true;
+      CDBDebug("No records in table found for dimension %s", dimName.c_str());
     }
 
     if (tableNotFound == false) {
       if (srvParams->isAutoLocalFileResourceEnabled() == true) {
         try {
-          CT::string databaseTime = *store->getRecord(0)->get(1);
+          CT::string databaseTime = store->records[0].get(1);
           if (databaseTime.length() < 20) {
             databaseTime.concat("Z");
           }
           databaseTime.setChar(10, 'T');
 
-          const auto fileDate = getFileDate(store->getRecord(0)->get(0)->c_str());
+          const auto fileDate = getFileDate(store->records[0].get(0).c_str());
 
           if (databaseTime.equals(fileDate) == false) {
             CDBDebug("Table was found, %s ~ %s : %d", fileDate.c_str(), databaseTime.c_str(), databaseTime.equals(fileDate));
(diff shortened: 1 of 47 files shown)


adagucserverEC/utils/XMLGenUtils.cpp
```diff
@@ -263,12 +263,12 @@ LayerMetadataDim handleMultipleValueDim(CDataSource *dataSource, CServerConfig::
...
-        queryValues.push_back(makeIsoStringFromDbString(*(values->getRecord(j)->get(0))));
+      for (size_t j = 0; j < values->records.size(); j++) {
+        queryValues.push_back(makeIsoStringFromDbString(values->records[j].get(0)));
       }
     } else {
-      for (size_t j = 0; j < values->size(); j++) {
-        queryValues.push_back(*(values->getRecord(j)->get(0)));
+      for (size_t j = 0; j < values->records.size(); j++) {
+        queryValues.push_back(values->records[j].get(0));
       }
```
mgrunbauer: For these kind of for loops, where we loop over `->records.size()`, is it possible to do something like?
```c++
for (auto& record : values->records) {
    queryValues.push_back(makeIsoStringFromDbString(record.get(0));
}
```

If you search for `->records.size()` there are a couple of examples in this PR that maybe could be written a bit nicer.

adagucserverEC/utils/UpdateLayerMetadata.cpp
```diff
@@ -23,10 +23,10 @@ int updateLayerMetadata(CRequest &request) {
   std::set<DatasetAndLayerPair> dataSetConfigsWithLayers;
 
   for (auto &dataset : datasetList) {
-    CT::string datasetAsCTString = dataset.c_str();
+    CT::string datasetAsCTString = dataset;
     CT::string baseDataSetNameCT = CT::basename(datasetAsCTString);
     baseDataSetNameCT.replaceSelf(".xml", "");
-    std::string datasetBaseName = baseDataSetNameCT.c_str();
+    std::string datasetBaseName = baseDataSetNameCT;
```
mgrunbauer: We're still using `CT::string` here, I think this could be refactored into:
```c++
std::string datasetBaseName = CT::replace(CT::basename(dataset), ".xml", "");
```

adagucserverEC/utils/XMLGenUtils.cpp
maartenplieger: Done!

adagucserverEC/utils/UpdateLayerMetadata.cpp
maartenplieger: Done!

adagucserverEC/utils/LayerMetadataStore.cpp
```diff
@@ -120,9 +120,9 @@ CT::string getLayerMetadataFromDb(MetadataLayer *metadataLayer, CT::string metad
   if (layerMetaDataStore == nullptr) {
     return "";
   }
-  auto records = layerMetaDataStore->getRecords();
+  auto &records = layerMetaDataStore->records;
   for (auto record: records) {
-    if (record.get("layername")->equals(layerName) && record.get("metadatakey")->equals(metadataKey)) {
+    if (record.get("layername") == layerName.c_str() && record.get("metadatakey") == metadataKey.c_str()) {
```
mgrunbauer: In `adagucserverEC/utils/LayerMetadataToJson.cpp` line 14, we do `record.get("layername") == layerName`.

I think this is the same `layerMetaDataStore->records`, why do we need `layerName.c_str()` here?

maartenplieger: Because this is still a CT::string and that has no == operator. But I can convert that too to std::string...

adagucserverEC/CRequest.cpp
```diff
@@ -595,21 +594,21 @@ int CRequest::queryDimValuesForDataSource(CDataSource *dataSource, CServerParams
       throw ServiceExceptionType::InvalidDimensionValue;
     }
 
-    for (size_t k = 0; k < store->getSize(); k++) {
-      CDBStore::Record *record = store->getRecord(k);
+    for (size_t k = 0; k < store->records.size(); k++) {
+      const CDBStore::Record *record = &store->records[k];
```
mgrunbauer: You could use the `for (auto record& : records)` here too.

The only thing we'd then lose by getting rid of the `k`, is the `CDBDebug`.

adagucserverEC/utils/LayerMetadataStore.cpp
maartenplieger: This is now converted.

hclasses/CDirReader.cpp
```diff
@@ -390,7 +359,7 @@ std::string getFileDate(const std::string &fileName) {
 
   std::map<std::string, std::string>::iterator it = lookupTableFileModificationDateMap.find(fileName);
   if (it != lookupTableFileModificationDateMap.end()) {
-    CT::string filemoddate = (*it).second.c_str();
+    std::string filemoddate = (*it).second.c_str();
     return filemoddate;
   }
```
mgrunbauer, line 357: You can also do:
```
...
if (it != lookupTableFileModificationDateMap.end()) {
    return it->second;
}
```

adagucserverEC/CDBAdapterPostgreSQL.cpp
```diff
@@ -524,46 +507,46 @@ int CDBAdapterPostgreSQL::autoUpdateAndScanDimensionTables(CDataSource *dataSour
...
       if (srvParams->isAutoLocalFileResourceEnabled() == true) {
         try {
-          CT::string databaseTime = store->getRecord(0)->get(1);
+          std::string databaseTime = store->records[0].values.at(1);
           if (databaseTime.length() < 20) {
-            databaseTime.concat("Z");
+            databaseTime += "Z";
           }
-          databaseTime.setChar(10, 'T');
+          databaseTime[10] = 'T';
```
mgrunbauer: The previous `setChar` did: `(location < length())`. Right now, if `databaseTime` is for whatever reason a small string, this results in undefined behaviour.

adagucserverEC/CRequest.cpp
maartenplieger: Done

hclasses/CDirReader.cpp
maartenplieger: Changed.

adagucserverEC/CDBAdapterPostgreSQL.cpp
maartenplieger: Good catch, how did you even spot this. Fixed

mgrunbauer (approved):

PR #729, opened 2026-07-01
Use binary search for shadeinterval in bilinear interpolation

mgrunbauer:
Profiling the generic warper with `interpolationmethod="bilinear"` for a dataset with many `<ShadeInterval>` elements, revealed that over 100 calls, 45% of runtime was spent in `memoizedDeterminePixelColorFromValue<float>`.

Further debugging showed adaguc eventually calls `determinePixelColorFromValue` from `CDrawFunction.cpp` for every pixel. Adaguc then loops through the shaded intervals, which is slow.

This PR uses `std::upper_bound` which uses binary search to find the matching shaded interval.

It helps especially with a large number of shaded intervals (e.g. 100), it's just about equal for 5 shaded intervals according to the benchmark (80 calls):

|Number of shaded intervals|Master|Branch|
|-|-|-|
|100|12.47|7.22|
|5|6.85|6.83|

Flamegraph master:
master
Flamegraph branch:
branch

PR #732, opened 2026-07-06
Vr8.0.0

maartenplieger:
**Version 8.0.0**

* Refactored makeIntervalFromTimeList by @maartenplieger in https://github.com/KNMI/adaguc-server/pull/726
* Refactor cdbstore by @maartenplieger in https://github.com/KNMI/adaguc-server/pull/728
* Use binary search for shadeinterval in bilinear interpolation by @mgrunbauer in https://github.com/KNMI/adaguc-server/pull/729
* Support inverted BBOX axis order for EPSG:32661 in WMS 1.3.0 by @maartenplieger in https://github.com/KNMI/adaguc-server/pull/731 - This can be a breaking change for clients. Note, EPSG:5041 is the same as EPSG:32661 but with the expected BBOX order of X,Y,X,Y and can be used instead.
* Update python libs, e.g. starlette from 1.1.0 to 1.3.1


doc/configuration/Projection.md
```diff
@@ -11,27 +11,36 @@ define the region corresponding to the projection.
     http://spatialreference.org/ref/epsg/4326/
 -   proj4 - The proj4 string corresponding to the identifier.
     https://trac.osgeo.org/proj/
+-   invertxyforwms130 -  for WMS 1.3.0, certain projections, the BBOX parameter requires to have order Y,X,Y,X instead of X,Y,X,Y This can be configured in the Projection element by setting invertxyforwms130 to true.
```
lukas-phaf: Instead, write that "WMS 1.3.0 strictly enforce authority-defined axis order"? Which means that for some projection the BBOX order is `Y,X,Y,X`?

Why do we have to do this by hand, BTW? Isn't the authority-defined axis order already handled correctly for EPSG 4326 when using parameters that are not in a BBOX?

NEWS.md
```diff
@@ -1,3 +1,12 @@
+
+**Version 8.0.0**
+
+* Refactored makeIntervalFromTimeList by @maartenplieger in https://github.com/KNMI/adaguc-server/pull/726
+* Refactor cdbstore by @maartenplieger in https://github.com/KNMI/adaguc-server/pull/728
+* Use binary search for shadeinterval in bilinear interpolation by @mgrunbauer in https://github.com/KNMI/adaguc-server/pull/729
+* Support inverted BBOX axis order for EPSG:32661 in WMS 1.3.0 by @maartenplieger in https://github.com/KNMI/adaguc-server/pull/731 - This can be a breaking change for clients. Note, EPSG:5041 is the same as EPSG:32661 but with the expected BBOX order of X,Y,X,Y and can be used instead.
```
lukas-phaf: Is "Support" the right word?

doc/configuration/Projection.md
```diff
@@ -11,27 +11,36 @@ define the region corresponding to the projection.
...
+  <Projection id="EPSG:50001" proj4="+proj=stere +lat_ts=60.0 +lat_0=90 +lon_0=-111.0 +k_0=1.0 +x_0=3020946.0 +y_0=7622187.0 +units=m +ellps=WGS84 +datum=WGS84 +no_defs ">
+    <LatLonBox minx="-2000000" miny="-2000000" maxx="10000000" maxy="8500000"/>
+  </Projection>
+  <Projection id="EPSG:54030" proj4="+proj=robin +lon_0=0 +x_0=0 +y_0=0 +ellps=WGS84 +datum=WGS84 +units=m +no_defs "/>
+  <Projection id="EPSG:32661" invertxyforwms130="true" proj4="+proj=stere +lat_0=90 +lat_ts=90 +lon_0=0 +k=0.994 +x_0=2000000 +y_0=2000000 +ellps=WGS84 +datum=WGS84 +units=m +no_defs"/>
+  <Projection id="EPSG:40000" proj4="+proj=stere +ellps=WGS84 +lat_0=90 +lon_0=0 +no_defs"/>
+  <Projection id="EPSG:900913" proj4=" +proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0 +k=1.0 +units=m +nadgrids=@null +no_defs"/>
+  <Projection id="EPSG:3067" proj4="+proj=utm +zone=35 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs +type=crs"/>
+  <Projection id="EPSG:3035" proj4="+proj=laea +lat_0=52 +lon_0=10 +x_0=4321000 +y_0=3210000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs +type=crs"/>
```
lukas-phaf, line 43: I looked up a random from this list on EPSG, and it also has N,E order.

I know this is just an example config file, but it is a copy of the PRD one, right? Maybe check them all?

doc/configuration/Projection.md
maartenplieger: Changed.


-->Why do we have to do this by hand, BTW? 
This is currently a server setting via the file `data/config/includes/Projection.include.xml`. The configuration is only once and is part of the repo.

NEWS.md
maartenplieger: Changed

doc/configuration/Projection.md
maartenplieger: Ouch, If so we with WMS 1.3.0, also need to change more stuff in the clients. The issue was specifically about EPSG:32661.

maartenplieger: I checked them all, EPSG:3035 is indeed with the Y,X order. Changed it here, we will have to update adaguc-viewer too.

maartenplieger: Updated the tests.

doc/configuration/Projection.md
```diff
@@ -11,27 +11,36 @@ define the region corresponding to the projection.
     http://spatialreference.org/ref/epsg/4326/
 -   proj4 - The proj4 string corresponding to the identifier.
     https://trac.osgeo.org/proj/
+-   invertxyforwms130 -  WMS 1.3.0 strictly enforces authority-defined axis order. Which means that for some projection the BBOX order is Y,X,Y,X?. Set `invertxyforwms130` to `true` to enforce this.
```
lukas-phaf:
```suggestion
-   invertxyforwms130 -  WMS 1.3.0 strictly enforces authority-defined axis order. Which means that for some projection the BBOX order is Y,X,Y,X. Set `invertxyforwms130` to `true` to enforce this.
```

tests/expectedoutputs/TestConvertLatLonBnds/test_ConvertLatLonBnds_getCapabilities.xml
```diff
@@ -81,7 +81,7 @@
 <BoundingBox CRS="EPSG:25831" minx="500000.000000" miny="5649824.888132" maxx="780631.054591" maxy="5879757.629488" />
 <BoundingBox CRS="EPSG:25832" minx="79142.928873" miny="5651728.682548" maxx="365786.750925" maxy="5889126.580102" />
 <BoundingBox CRS="EPSG:28992" minx="-12557.532230" miny="334483.538071" maxx="268209.963283" maxy="559655.555150" />
-<BoundingBox CRS="EPSG:3035" minx="3830472.135998" miny="3103060.820291" maxx="4119643.144373" maxy="3343905.347139" />
+<BoundingBox CRS="EPSG:3035" minx="3103060.820291" miny="3830472.135998" maxx="3343905.347139" maxy="4119643.144373" />
```
lukas-phaf, line 84: I am sure I mentioned this before, but this is horrible, `minx` and then a `Y` value is listed. The attributes should really be renamed. But not now...

maartenplieger: This is according to the WMS 1.3.0 standard. It is confusing indeed.

lukas-phaf (approved):

PR #737, opened 2026-07-08
Changed CT::string to std::string for hclasses directory

maartenplieger:
- More refactoring of CT::string -> std::string

### DIFF
--- a/CCDFDataModel/CProj4ToCF.cpp
+++ b/CCDFDataModel/CProj4ToCF.cpp
@@ -35,14 +35,14 @@ float CProj4ToCF::CProj4ToCF::convertToM(float fValue) {
 
 CT::string CProj4ToCF::getProj4Value(const char *proj4Key, std::vector<CKeyValuePair> projKVPList) {
   for (size_t j = 0; j < projKVPList.size(); j++) {
-    if (projKVPList[j].key.equals(proj4Key) == true) {
+    if (projKVPList[j].key == proj4Key) {
       return projKVPList[j].value;
     }
   }
   throw(0);
 }
 
-float CProj4ToCF::getProj4ValueF(const char *proj4Key, std::vector<CKeyValuePair> projKVPList, float defaultValue, float((*conversionfunction)(float))) {
+float CProj4ToCF::getProj4ValueF(const char *proj4Key, std::vector<CKeyValuePair> projKVPList, float defaultValue, float(*conversionfunction)(float)) {
   float value = defaultValue;
   try {
     value = getProj4Value(proj4Key, projKVPList).toFloat();
@@ -414,14 +414,14 @@ int CProj4ToCF::convertProjToCF(CDF::Variable *projectionVariable, const char *p
   try {
 
     for (size_t j = 0; j < projKVPList.size(); j++) {
-      if (projKVPList[j].key.equals("proj")) {
-        if (projKVPList[j].value.equals("stere")) {
+      if (projKVPList[j].key == "proj") {
+        if (projKVPList[j].value == "stere") {
           initStereoGraphic(projectionVariable, projKVPList);
           foundProj = 1;
         }
-        if (projKVPList[j].value.equals("geos")) {
+        if (projKVPList[j].value == "geos") {
           for (size_t j2 = 0; j2 < projKVPList.size(); j2++) {
-            if (projKVPList[j2].key.equals("height_from_earth_center")) {
+            if (projKVPList[j2].key == "height_from_earth_center") {
               initMSGPerspective(projectionVariable, projKVPList);
               foundProj = 1;
             }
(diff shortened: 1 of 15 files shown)


hclasses/CReporter.cpp
```diff
@@ -26,7 +26,7 @@ CReporter *CReporter::getInstance() {
   return instance;
 }
 
-void CReporter::addMessage(const CT::string message, CReportMessage::Severities severity, CReportMessage::Categories category, CT::string documentationLink, const char *file, int line,
+void CReporter::addMessage(const std::string message, CReportMessage::Severities severity, CReportMessage::Categories category, std::string documentationLink, const char *file, int line,
```
mgrunbauer: Can it be `const std::string &message` and `const std::string &documentationLink`?

Different question: when is this `addMessage` used, can this not use the "regular" logging using `CDBDebug` and `CDBError`?

hclasses/CReporter.cpp
```diff
@@ -37,7 +37,7 @@ void CReporter::addMessage(const CT::string message, CReportMessage::Severities
   messageList.push_back(_message);
 }
 
-void CReporter::writeMessageToLog(const CT::string message, CReportMessage::Severities severity, const char *file, int line, const char *className) const {
+void CReporter::writeMessageToLog(const std::string message, CReportMessage::Severities severity, const char *file, int line, const char *className) const {
```
mgrunbauer: `const std::string &message`?

hclasses/CReporter.cpp
```diff
@@ -57,6 +57,6 @@ void CReporter::writeMessageToLog(const CT::string message, CReportMessage::Seve
 
 const std::list<CReportMessage> CReporter::getMessageList() const { return messageList; }
 
-void CReporter::filename(const CT::string filename) { this->_filename = filename; }
+void CReporter::filename(const std::string filename) { this->_filename = filename; }
```
mgrunbauer: `const std::string &filename`?

CCDFDataModel/CProj4ToCF.cpp
```diff
@@ -414,14 +414,14 @@ int CProj4ToCF::convertProjToCF(CDF::Variable *projectionVariable, const char *p
   try {
 
     for (size_t j = 0; j < projKVPList.size(); j++) {
-      if (projKVPList[j].key.equals("proj")) {
-        if (projKVPList[j].value.equals("stere")) {
+      if (projKVPList[j].key == "proj") {
```
mgrunbauer: Should we use the for auto?:
```
for (const auto &kvp : projKVPList) {
  if (kvp.key == "proj") {
    ...
```

CCDFDataModel/CReadPNG.cpp
```diff
@@ -108,8 +107,7 @@ CPNGRaster *CReadPNG_read_png_file(const char *file_name, bool pngReadHeaderOnly
 
   png_read_info(png_ptr, info_ptr);
 
-  CPNGRaster *pngRaster = new CPNGRaster();
-
+  CPNGRaster *volatile pngRaster = new CPNGRaster();
```
mgrunbauer: What does the `*volatile` do here, why is it needed?

adagucserverEC/CConvertADAGUCPoint.cpp
```diff
@@ -858,7 +859,7 @@ int CConvertADAGUCPoint::convertADAGUCPointData(CDataSource *dataSource, int mod
               }
               /* Get the last pushed point from the array and push the character text data in the paramlist */
               const char *b = ((const char **)pointVar[d]->data)[pPoint];
-              lastPoint->paramList.push_back({.key = key, .description = description, .value = b});
+              lastPoint->paramList.push_back({.key = key, .description = description, .value = CT::fromCharPointer(b)});
```
mgrunbauer, line 862: Why not just `.value = std::string(b)`?

Is this because `b` might be a nullptr?

hclasses/CReporter.cpp
maartenplieger: Yes I think so. But this is functionality made by the KDP team. I rather leave it like it is for the time being.

I changed to `const std::string &` where applicable

hclasses/CReporter.cpp
maartenplieger: Yes!

adagucserverEC/CConvertADAGUCPoint.cpp
maartenplieger: Yes, assigning a nullptr to a std::string is undefined behaviour. CT::string `=` operator has a check for this. That is why we need this helper function.

CCDFDataModel/CReadPNG.cpp
maartenplieger: the compiler tries to optimize and you get a very weird error if you don't say it is volatile:

```
CReadPNG.cpp:110:15: error: variable ‘pngRaster’ might be clobbered by ‘longjmp’ or ‘vfork’
```
This has to do with the -O2 flag and it seems to be an issue with gcc 13.3.0...

CCDFDataModel/CProj4ToCF.cpp
maartenplieger: Changed this, much cleaner

hclasses/CReporter.cpp
maartenplieger: Fixed

CCDFDataModel/CReadPNG.cpp
maartenplieger: More info here https://github.com/wahern/hexdump/issues/1

maartenplieger: I am going to revert the volatile keyword and just disable the warning, will mention a todo.

maartenplieger: Done.

mgrunbauer (approved):

Issue #674, opened 2026-03-30
Check why 10 minute observations render slowly


maartenplieger:
<img width="731" height="620" alt="Image" src="https://github.com/user-attachments/assets/beffe6ea-f98d-4308-ac20-9634a6333195" />

maartenplieger:
Improvements are being made in https://github.com/KNMI/adaguc-server/pull/738

PR #738, opened 2026-07-09
Improve point render speed

maartenplieger:
- Refactored CT::string to std::string in CConvertADAGUCPoint.cpp and CImgRenderPoints.cpp
- Optimizing point rendering code 
- Added helper to the CCDF API functions so CConvertADAGUCPoint could be simplified
- Avoid using creating a float grid with the size of the getmap view as this is not needed for point data ((dataSource->hasFieldData) 
- Remove station id from timeseries

This is for https://github.com/KNMI/adaguc-server/issues/674


adagucserverEC/CConvertADAGUCPoint.cpp
```diff
@@ -24,13 +24,14 @@
...
 #include "CConvertADAGUCPoint.h"
+#include <algorithm>
 #include "CFillTriangle.h"
 #include "CImageWarper.h"
 #include "CConvertADAGUCPoint_convert_BIRA_IASB_NETCDF.cpp"
 #include "utils/minMax.h"
 #include "CStyleConfiguration.h"
 #include "CTString.h"
-
+#define MEASURETIME
```
maartenplieger: Remove this

ernstdevreede (approved):

PR #755, opened 2026-09-08
Support showinlegend property in ShadeInterval

### LINKED ISSUE #754: Add a property to ShadeInterval to indicate whether it should be listed in the legendgraphic
When the property is set to `displayInLegendGraphic="false"`, the ShadeInterval item should not be rendered in the getlegendgraphic legend.

maartenplieger:
- Support showinlegend property in ShadeInterval to filter out items for the GetLegendGraphic request: https://github.com/KNMI/adaguc-server/issues/754

Closes  https://github.com/KNMI/adaguc-server/issues/754


adagucserverEC/CLegendRenderers/CCreateLegendRenderDiscreteLegend.cpp
```diff
@@ -24,11 +24,9 @@ void plotNumericLabels(CDrawImage *legendImage, double scaling, const std::strin
   float numericMinVal = atof(s.attr.min.c_str());
 
   // Calculate number of decimals for min column
-  CT::string floatFormatMin;
-  floatFormatMin.print("%%.%df", maxDecimalWidth(minColumn));
+  std::string floatFormatMin = CT::printf("%%.%df", maxDecimalWidth(minColumn));
 
-  CT::string tempText;
-  tempText.print(floatFormatMin.c_str(), numericMinVal);
+  std::string tempText = CT::printf(floatFormatMin.c_str(), numericMinVal);
```
mgrunbauer: I just discovered that printf can already do dynamic precision stuff with the `*` symbol:
```
std::string tempText = CT::printf("%.*f", maxDecimalWidth(minColumn), numericMinVal);
```
That would save us an extra call to `CT::printf`.

The same can be done for the `maxColumn`.

adagucserverEC/CLegendRenderers/CCreateLegendRenderDiscreteLegend.cpp
```diff
@@ -24,11 +24,9 @@ void plotNumericLabels(CDrawImage *legendImage, double scaling, const std::strin
   float numericMinVal = atof(s.attr.min.c_str());
```
mgrunbauer: This file does a lot of double `atof(s.attr.min.c_str()`.

For e.g. `<Vector>` elements we decided to store min/max attributes as `double` directly in the `XMLE_ShadeInterval` struct. That means doing the conversion once, which can then clean up some extra atof's here. What do you think, is that possible or tricky?

We might need both representations (double and std:string) because we're also using these fields for formatting I think.

adagucserverEC/CServerConfig_CPPXSD.h
```diff
@@ -401,6 +401,7 @@ struct CServerConfig : CXMLObjectInterface {
   struct XMLE_ShadeInterval : CXMLObjectInterface {
     struct Cattr {
       std::string min, max, label, fillcolor, bgcolor;
+      bool showInLegend = true;
```
mgrunbauer: I think all XML attributes are currently lowercase. With longer words we either put everything lowercase (as in: `textoutlinecolor`) or sometimes we add a underscore (as in: `standard_name`).

I think we should be consistent and keep everything lowercase. I would then prefer `showinlegend`.

adagucserverEC/CLegendRenderers/CCreateLegendRenderDiscreteLegend.cpp
```diff
@@ -240,26 +239,26 @@ int CCreateLegend::renderDiscreteLegend(CDataSource *dataSource, CDrawImage *leg
...
     }
 
     // Only a subset of intervals to appear on screen to save space
     size_t drawIntervals = maxInterval - minInterval;
 
     // Calculate columns and text properties
-    std::vector<CT::string> minColumn = extractColumn(drawIntervals, minInterval, styleConfiguration->shadeIntervals, true);
-    std::vector<CT::string> maxColumn = extractColumn(drawIntervals, minInterval, styleConfiguration->shadeIntervals, false);
+    std::vector<CT::string> minColumn = extractColumn(drawIntervals, minInterval, shadeIntervalsFilteredForLegendGraphic, true);
+    std::vector<CT::string> maxColumn = extractColumn(drawIntervals, minInterval, shadeIntervalsFilteredForLegendGraphic, false);
```
mgrunbauer: I think if you change `extractColumn` to return `std::vector<std::string>` (should be ok, the attributes are already `std::string`), you've completely eliminated `CT::string` from this whole file... :)

adagucserverEC/CLegendRenderers/CCreateLegendRenderDiscreteLegend.cpp
```diff
@@ -24,11 +24,9 @@ void plotNumericLabels(CDrawImage *legendImage, double scaling, const std::strin
   float numericMinVal = atof(s.attr.min.c_str());
 
   // Calculate number of decimals for min column
-  CT::string floatFormatMin;
-  floatFormatMin.print("%%.%df", maxDecimalWidth(minColumn));
+  std::string floatFormatMin = CT::printf("%%.%df", maxDecimalWidth(minColumn));
```
mgrunbauer: We're also doing multiple calls to `maxDecimalWidth` for minColumn and maxColumn. I think we can do these calls one at the beginning and reuse the value.

Note that `maxDecimalWidth` (and all related methods in `numericutils.cpp` passes the vector by value, maybe change these to references too?

adagucserverEC/CServerConfig_CPPXSD.h
maartenplieger: Good point, changed to showinlegend.

adagucserverEC/CLegendRenderers/CCreateLegendRenderDiscreteLegend.cpp
maartenplieger: Nice feature!

adagucserverEC/CLegendRenderers/CCreateLegendRenderDiscreteLegend.cpp
maartenplieger: I tried to convert this in the XMLE_ShadeInterval, it causes many changes. But I guess it is worth it.

adagucserverEC/CLegendRenderers/CCreateLegendRenderDiscreteLegend.cpp
maartenplieger: Changed a lot more like these, also used const & where possible.

adagucserverEC/CLegendRenderers/CCreateLegendRenderDiscreteLegend.cpp
maartenplieger: Good suggestion, did this!

adagucserverEC/CLegendRenderers/numericutils.cpp
```diff
@@ -27,22 +27,22 @@ int maxIntWidth(std::vector<CT::string> column) {
...
   return isNeg;
 }
 
-int maxDecimalWidth(std::vector<CT::string> column) {
+int maxDecimalWidth(const std::vector<std::string> &column) {
   int maxDecimals = 0;
 
-  for (auto &item : column) {
-    int dotIndex = item.indexOf(".");
+  for (std::string item: column) {
```
mgrunbauer: Should make this `for (const std::string  &item : column)`, otherwise it still copies.

adagucserverEC/CLegendRenderers/CCreateLegendRenderContinuousLegend.cpp
```diff
@@ -309,9 +304,9 @@ int CCreateLegend::renderContinuousLegend(CDataSource *dataSource, CDrawImage *l
     }
   }
   // Get units
-  CT::string units;
+  std::string units;
   if (dObjgetUnits(*dataSource->getDataObject(0)).length() > 0) {
-    units.concat(dObjgetUnits(*dataSource->getDataObject(0)).c_str());
+    units += (dObjgetUnits(*dataSource->getDataObject(0)).c_str());
```
mgrunbauer: In `CCreateLegendRenderDiscreteLegend.cpp` you did this:
```c++
std::string units = dObjgetUnits(*dataSource->getDataObject(0));
```

I think you can do that here too?

adagucserverEC/CLegendRenderers/CCreateLegendRenderDiscreteLegend.cpp
mgrunbauer: Nice, I think it makes sense to do the convert early. I like it! 👍

adagucserverEC/CLegendRenderers/numericutils.cpp
maartenplieger: Changed. (Sharp!)

adagucserverEC/CLegendRenderers/CCreateLegendRenderContinuousLegend.cpp
maartenplieger: Changed

mgrunbauer (approved):

PR #756, opened 2026-09-09
Removed ct::string class

maartenplieger:
- class CT::string is now entirely replaced with std::string
- const char pointers are replaced with std::string where possible
- Restructured includes
- Replaced debug macros with static booleans for better readability
- Removed dead code and unused files.
- Moved implementation logic from h to cpp files
- Removed comment code

No changed functionality.


adagucserverEC/CConvertKNMIH5EchoToppen.cpp
mgrunbauer, line 1: You added:
```
 CDBDebug("String [%s]\n---\n%s\n ---", projectionString.c_str(), CDF::dump(cdfObject0).c_str());
```

Was this temporarily? Did you want to leave it in?

adagucserverEC/anydump.cpp
mgrunbauer, line 1: This file still calls CT::string methods like `.equals` which don't exist on `std::string`.

I think we don't do anything with these files, we should either fix the `.equals` or remove if these files are deprecated.

adagucserverEC/aggregate_dim.cpp
mgrunbauer, line 1: This file still calls CT::string methods like `.endsWith` which don't exist on `std::string`.

I think we don't do anything with this file, we should either fix the `.equals` or remove if these files are deprecated.

adagucserverEC/CConvertKNMIH5EchoToppen.cpp
maartenplieger: How did you even manage to spot this. Yes I left this accidentally in, good find!

adagucserverEC/anydump.cpp
maartenplieger: I fixed anydump, it is now in the bin folder. It is actually a useful little tool.

adagucserverEC/aggregate_dim.cpp
maartenplieger: I think this is indeed deprecated. Removed.

mgrunbauer (approved):

lukas-phaf (approved):
Unreviewable, but it would be the same if the CT::string replacement was done by hand.

PR #760, opened 2026-09-14
Fixes needed to compile using clang on aarch64

mgrunbauer:
Could not compile adaguc using clang on aarch64. 3 issues:

- `-Wclobbered` is gcc only.
- `-Wno-restrict` is gcc only
- We're formatting `int64_t` using %ld, which appears to work under amd64/gcc, but fails under aarch64/clang. The new `CT::printf` has stricter (more correct?) formatting rules.
  - A few cases in GeoJSON-related code and in CCDFHDF5IO now use the `PRId64` macro from `cinttypes`.


adagucserverEC/CConvertGeoJSON.cpp
```diff
@@ -586,7 +589,7 @@ void CConvertGeoJSON::getBBOX(CDFObject *, BBOX &bbox, json_value &json, std::ve
           if (id.type == json_string) {
             featureId = id.u.string.ptr;
           } else if (id.type == json_integer) {
-            featureId = CT::printf("%1ld", id.u.integer);
+            featureId = CT::printf("%" PRId64, id.u.integer);
```
maartenplieger, line 592: Had to look this up.

maartenplieger (approved):

PR #772, opened 2026-09-15
Avoid using new/delete for the config parser

maartenplieger:
- Avoid using heap allocated configuration elements in configuration vectors. No more new/delete here.
- Store configuration objects by value rather than manually allocating and freeing them on the heap.
- Macros XMLE_ADDOBJ, XMLE_SETOBJ and XMLE_DELOBJ are removed from adagucserverEC/CXMLSerializerInterface.h
- Instead, use methods addXmlObj and setXmlObj


adagucserverEC/CAutoConfigure.cpp
```diff
@@ -84,12 +84,11 @@ int CAutoConfigure::autoConfigureDimensions(CDataSource *dataSource) {
     try {
 
       for (auto &record: store->records) {
-        CServerConfig::XMLE_Dimension *xmleDim = new CServerConfig::XMLE_Dimension();
+        CServerConfig::XMLE_Dimension *xmleDim = addXmlObj(dataSource->cfgLayer->Dimension);
```
maartenplieger, line 87: Need to check this carefully

adagucserverEC/CAutoResource.cpp
```diff
@@ -521,39 +518,33 @@ void CAutoResource::addXMLLayerToConfig(CServerParams *const srvParam, CDFObject
...
-        xmleImageText->attr.attribute = srvParam->cfg->AutoResource[0]->ImageText[0]->attr.attribute;
+      if (srvParam->cfg->AutoResource[0].ImageText[0].attr.attribute.empty() == false) {
+        xmleImageText->attr.attribute = srvParam->cfg->AutoResource[0].ImageText[0].attr.attribute;
       }
     }
   }
 
   // Add the layer to the configuration
-  srvParam->cfg->Layer.push_back(xmleLayer);
+  srvParam->cfg->Layer.push_back(std::move(xmleLayer));
```
maartenplieger, line 549: Check

adagucserverEC/CXMLSerializerInterface.h
mgrunbauer, line 77: Now that XML nodes are stored directly in the vectors, a vector reallocation (growing vector capacity) must relocate the existing nodes. Previously, this only required copying or moving pointers.

I think generally this PR is better overall (no new/delete etc), and I think that with our configs, the extra moving is probably not noticeable.

If we want to avoid it, we could explicitly define `noexcept` move operations for the base `struct CXMLObjectInterface`. Currently, the compiler cannot generate this move constructor because we define a virtual destructor.

adagucserverEC/utils/ConfigurationUtils.cpp
```diff
@@ -18,13 +18,13 @@ std::vector<std::string> getEnabledDatasetsConfigurations(CServerParams *srvPara
     CDBWarning("No dataset paths are configured");
   }
   for (auto dataset: srvParam->cfg->Dataset) {
```
mgrunbauer: Can be `const auto &dataset`

adagucserverEC/utils/lintDataset.cpp
```diff
@@ -85,22 +85,22 @@ int lintDataset(std::string config) {
   }
   // Remove the ones already configured in the base
   for (auto Legend: srvParamsBase.cfg->Legend) {
-    removeStringFromVector(Legend->attr.name, availableLegendNames);
+    removeStringFromVector(Legend.attr.name, availableLegendNames);
   }
 
   for (auto Style: srvParamsBase.cfg->Style) {
```
mgrunbauer: Can be `const auto &style`

adagucserverEC/utils/lintDataset.cpp
```diff
@@ -85,22 +85,22 @@ int lintDataset(std::string config) {
   }
   // Remove the ones already configured in the base
   for (auto Legend: srvParamsBase.cfg->Legend) {
```
mgrunbauer: `const auto &legend`

adagucserverEC/utils/lintDataset.cpp
```diff
@@ -85,22 +85,22 @@ int lintDataset(std::string config) {
...
+    removeStringFromVector(Legend.attr.name, availableLegendNames);
   }
 
   for (auto Style: srvParamsBase.cfg->Style) {
-    removeStringFromVector(Style->attr.name, availableStyleNames);
+    removeStringFromVector(Style.attr.name, availableStyleNames);
   }
 
   // Remove the ones with standardname mappings, these cannot be properly checked.
   for (auto Style: srvParamsBase.cfg->Style) {
```
mgrunbauer: `const auto &style`

mgrunbauer (dismissed):

adagucserverEC/CXMLSerializerInterface.h
maartenplieger: Excellent remark. I was not aware of this. Added noxecept = default.

adagucserverEC/utils/lintDataset.cpp
maartenplieger: Changed this everywhere in the code to this pattern using clang-tidy.

hclasses/CDirReader.cpp
```diff
@@ -51,12 +51,12 @@ const std::vector<std::string> CDirReader::listDir(const char *directory, bool r
     /* print all the files and directories within directory */
     while ((ent = readdir(dir)) != NULL) {
       std::string fullName = directory;
-      fullName += fullName + "/" + ent->d_name;
+      fullName = fullName + "/" + ent->d_name;
```
mgrunbauer, line 54: Did nothing break because of this? Good that you caught it! 👍

mgrunbauer (approved):

hclasses/CDirReader.cpp
maartenplieger: No, and it is a real bug!

adagucserverEC/CXMLSerializerInterface.h
lukas-phaf: Goed punt, maar ik vond het eigenlijk wel een interessante vraag wat move versus copy nu precies oplevert voor een object met alleen een default constructor.
Voor het XML object zelf verandert er volgens mij niks, maar het verschil is dat bijvoorbeeld std:string's die er in zitten nu ge"moved" kunnen worden, dus de data op de heap die zij gebruiken hoeft niet opnieuw gealloceerd te worden. Maar misschien was dit voor iedereen al obvious.

Goede verandering dus.

lukas-phaf: Claude komt nog wel met deze opmerking, heeft volgens mij wel een goed punt:
"Whether you want public copy and move on a polymorphic base is a separate question. Copying through a base reference slices off the derived part. The C++ Core Guidelines (C.67) recommend making them protected or = delete in polymorphic bases, unless you really do copy these objects by value somewhere."

Code wordt dan zoiets:
```
struct CXMLObjectInterface {
  std::string elementValue;

  virtual ~CXMLObjectInterface() = default;

  virtual CXMLObjectInterface *addElement(const std::string &) { return nullptr; }
  virtual void handleValue() {}
  virtual bool addAttribute(const attribute &) { return false; }

protected:
  CXMLObjectInterface() = default;
  CXMLObjectInterface(const CXMLObjectInterface &) = default;
  CXMLObjectInterface(CXMLObjectInterface &&) noexcept = default;
  CXMLObjectInterface &operator=(const CXMLObjectInterface &) = default;
  CXMLObjectInterface &operator=(CXMLObjectInterface &&) noexcept = default;
};
```

